PR #40, opened 2024-05-16
Feat/tedi 261 tedi components export

airikej:


libs/react-components/package.json
```diff
@@ -5,9 +5,13 @@
...
   "module": "index.js",
   "exports": {
+    "./tedi": {
+      "types": "./tedi/index.d.ts",
+      "import": "./index.js"
+    },
     "./community": {
       "types": "./community/index.d.ts",
-      "import": "./index.js"
+      "import": "./index2.js"
```
tobre6, line 14: index2 name is not very good approach. Can it be just index.js? Or otherwise pehaps community.js and tedi.js respectively.

tobre6 (changes requested):

tobre6 (approved):

Issue #45, opened 2024-05-20
Breadcrumbs: WCAG - last element is a link


kristiina-richards:
### Description of the problem
Last element in breadcrumbs looks like a text but is a link. 
**Requirement:** [link](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html)

PR #46, opened 2024-05-20
fix(Breadcrumbs): Make last element text

kristiina-richards:
https://github.com/TEHIK-EE/tedi-design-system/issues/45


kristiina-richards:
https://kristiina-richards.github.io/tedi-design-system/rc/?path=/docs/community-components-layout-breadcrumbs--docs

HarlesPilter (approved):

Issue #54, opened 2024-05-23
FileUpload: File size checking


MVaik:
Option to add a file size limit and a callback for any rejected files.

PR #55, opened 2024-05-23
feat(FileUpload): added optional size checking and callback

MVaik:
https://github.com/TEHIK-EE/tedi-design-system/issues/54

### DIFF
--- a/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
+++ b/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
@@ -1,8 +1,8 @@
 import { Meta, StoryFn, StoryObj } from '@storybook/react';
 import { ComponentProps, useState } from 'react';
 
-import Notification from '../../notification/notification';
 import { VerticalSpacing } from '../../vertical-spacing';
+import FormHelper from '../form-helper/form-helper';
 import { MultipleHandledTemplate } from './examples/multiple-handled';
 import FileUpload from './file-upload';
 
@@ -89,16 +89,14 @@ export const PdfAndTxtOnly: Story = {
 };
 
 const MaxSizeTemplate: StoryFn<ComponentProps<typeof FileUpload>> = (args) => {
-  const [invalidSizeFiles, setInvalidSizeFiles] = useState<File[]>([]);
-
+  const [helperText, setHelperText] = useState<string | undefined>();
+  const handleInvalidFiles = (files: File[]) => {
+    setHelperText(`File(s) ${files.map((file) => `'${file.name}'`).join(', ')} are too big`);
+  };
   return (
     <VerticalSpacing>
-      <FileUpload {...args} onInvalidSize={setInvalidSizeFiles} />
-      {invalidSizeFiles.length > 0 && (
-        <Notification type="error">
-          File(s) {invalidSizeFiles.map((file) => `'${file.name}'`).join(', ')} are too big
-        </Notification>
-      )}
+      <FileUpload {...args} onInvalidSize={handleInvalidFiles} onChange={() => setHelperText(undefined)} />
+      {helperText && <FormHelper text={helperText} type="error" />}
     </VerticalSpacing>
   );
 };
--- a/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
+++ b/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
@@ -95,8 +95,12 @@ const MaxSizeTemplate: StoryFn<ComponentProps<typeof FileUpload>> = (args) => {
   };
   return (
     <VerticalSpacing>
-      <FileUpload {...args} onInvalidSize={handleInvalidFiles} onChange={() => setHelperText(undefined)} />
-      {helperText && <FormHelper text={helperText} type="error" />}
+      <FileUpload
+        {...args}
+        onInvalidSize={handleInvalidFiles}
+        onChange={() => setHelperText(undefined)}
+        helper={helperText ? { text: helperText } : undefined}
+      />
     </VerticalSpacing>
   );
 };
--- a/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
+++ b/libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
@@ -1,5 +1,8 @@
-import { Meta, StoryObj } from '@storybook/react';
+import { Meta, StoryFn, StoryObj } from '@storybook/react';
+import { ComponentProps, useState } from 'react';
 
+import Notification from '../../notification/notification';
+import { VerticalSpacing } from '../../vertical-spacing';
 import { MultipleHandledTemplate } from './examples/multiple-handled';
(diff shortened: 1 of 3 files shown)


MVaik:
https://mvaik.github.io/tedi-design-system/rc/?path=/docs/community-components-form-fileupload--docs

kasparsuvi1:
Should we show default `type="help"` FormHelper when `size` or `accept` are provided? (I assume we would need it inside applications either way)
So when there is size/type restriction user have that info inside FormHelper. Passing custom helper property from application should overwrites this behavior - to show errors etc.

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
```diff
@@ -84,3 +84,12 @@ export const PdfAndTxtOnly: Story = {
     accept: '.pdf,.txt',
   },
 };
+
+export const SizeLimited: Story = {
```
kasparsuvi1, line 88: Example should show how the `onInvalidSize` fn can be used to notify user, when some files are not uploaded. But some are.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
```diff
@@ -69,6 +69,14 @@ export interface FileUploadProps extends FormLabelProps {
    * @default false
    */
   disabled?: boolean;
+  /**
+   * Size limit in bytes
+   */
+  sizeLimit?: number;
```
kasparsuvi1, line 75: Maybe we can use size in  MB instead of bytes here? So the applications can pass just `sizeLimit={5}`, not `sizelimit={5242880}` (maxAllowedSize = 5 * 1024 * 1024)

kasparsuvi1, line 75:
```suggestion
  maxSize?: number;
```
?

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
MVaik: Added an example.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
MVaik: Changed.

MVaik:
> Should we show default `type="help"` FormHelper when `size` or `accept` are provided? (I assume we would need it inside applications either way) So when there is size/type restriction user have that info inside FormHelper. Passing custom helper property from application should overwrites this behavior - to show errors etc.

Added a default helper, new line doesn't seem to be supported so I separated different texts with a comma.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
```diff
@@ -70,20 +70,33 @@ export interface FileUploadProps extends FormLabelProps {
...
+const getDefaultHelpers = (
+  { accept, maxSize }: Partial<FileUploadProps>,
+  getLabel: ILabelContext['getLabel']
+): FormHelperProps | undefined => {
+  const text = [
+    accept && `${getLabel('file-upload.accept')} ${accept.replace(',', ', ')}`,
+    maxSize && `${getLabel('file-upload.max-size')} ${maxSize}MB`,
+  ]
+    .filter(Boolean)
+    .join(', ');
```
kasparsuvi1, line 91: Because both translation start with capital letter, we should separate them with dot not comma.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
```diff
@@ -70,20 +70,33 @@ export interface FileUploadProps extends FormLabelProps {
...
    * Callback to be used when files are rejected due to size
    */
   onInvalidSize?: (files: File[]) => void;
 }
 
+const getDefaultHelpers = (
+  { accept, maxSize }: Partial<FileUploadProps>,
+  getLabel: ILabelContext['getLabel']
+): FormHelperProps | undefined => {
+  const text = [
```
kasparsuvi1, line 86: Return early undefined if accept and maxSize are not defined. So the text.length check is not needed.

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
```diff
@@ -85,11 +88,28 @@ export const PdfAndTxtOnly: Story = {
...
 };
 
+const MaxSizeTemplate: StoryFn<ComponentProps<typeof FileUpload>> = (args) => {
+  const [invalidSizeFiles, setInvalidSizeFiles] = useState<File[]>([]);
+
+  return (
+    <VerticalSpacing>
+      <FileUpload {...args} onInvalidSize={setInvalidSizeFiles} />
+      {invalidSizeFiles.length > 0 && (
+        <Notification type="error">
```
kasparsuvi1, line 98: Nice-to-have, but i think more "nice" and semantically correct is showing that error with fileUpload.helper error/important text.
Also when uploading new files, we should reset invalidSizeFiles. Because that error is not relevant anymore

kasparsuvi1 (changes requested):

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
MVaik: Switched to formhelper and reset on change.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
MVaik: Added early return.

libs/react-components/community/src/components/form/file-upload/file-upload.tsx
MVaik: Changed separator.

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
```diff
@@ -89,16 +89,14 @@ export const PdfAndTxtOnly: Story = {
...
   return (
     <VerticalSpacing>
-      <FileUpload {...args} onInvalidSize={setInvalidSizeFiles} />
-      {invalidSizeFiles.length > 0 && (
-        <Notification type="error">
-          File(s) {invalidSizeFiles.map((file) => `'${file.name}'`).join(', ')} are too big
-        </Notification>
-      )}
+      <FileUpload {...args} onInvalidSize={handleInvalidFiles} onChange={() => setHelperText(undefined)} />
+      {helperText && <FormHelper text={helperText} type="error" />}
```
kasparsuvi1, line 99: Should be used from FileUpload propertie `helper` not rendered separated.

kasparsuvi1 (changes requested):

libs/react-components/community/src/components/form/file-upload/file-upload.stories.tsx
MVaik: Done.

kasparsuvi1 (approved):

PR #58, opened 2024-05-27
fix(select): required props passed to input

kristiina-richards:


kasparsuvi1:
#49

kasparsuvi1 (approved):

kasparsuvi1:
Seems fine, but why the fork deploys and builds fail? Should we merge when PR fork actions are failing?
@tobre6 
https://github.com/kristiina-richards/tedi-design-system/actions/runs/9255717904

tobre6:
> Seems fine, but why the fork deploys and builds fail? Should we merge when PR fork actions are failing? @tobre6 https://github.com/kristiina-richards/tedi-design-system/actions/runs/9255717904

Kristiina already reached help for failing deploys if the fork repo uses branches. Will investigating the deploy today. Probably not related with the build.

Issue #51, opened 2024-05-21
ChoiceGroup: WCAG issues


kristiina-richards:
### Description of the problem

JAWS reads out checkbox and radio button labels incorrectly - first the checkbox and the the label. Should be read only once not as 2 separate items.  
**Requirement:** [link](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)

kasparsuvi1:
Breaks check/radio selection by click on the label. Change reverted

PR #59, opened 2024-05-28
fix(choice-group): fix radio and check button screenreader bug

kristiina-richards:
https://kristiina-richards.github.io/tedi-design-system/bug/choicegroup-wcag/?path=/docs/community-components-form-choicegroup--docs


kasparsuvi1:
#51

HarlesPilter (approved):

kasparsuvi1:
Breaks selecting Check/Radio by clickin on the label. Change reverted

Issue #52, opened 2024-05-23
Datepicker: WCAG issues - mobile view


kristiina-richards:
### Description of the problem

Datepicker component seems to have a mobile view but it should also be called when used in browser and zoomed in at 400%
**Requirement:** [link](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)

PR #60, opened 2024-05-28
fix(datepicker, datetimepicker, timepicker): fix mobile view for pickers in zoomed view

kristiina-richards:
https://kristiina-richards.github.io/tedi-design-system/fix/datepicker-mobile/?path=/docs/community-components-form-pickers-datepicker--docs

### DIFF
--- a/libs/react-components/community/src/components/form/pickers/datepicker/datepicker.tsx
+++ b/libs/react-components/community/src/components/form/pickers/datepicker/datepicker.tsx
@@ -1,8 +1,9 @@
-import { DatePicker as MuiDatePicker } from '@mui/x-date-pickers';
+import { DesktopDatePicker, MobileDatePicker } from '@mui/x-date-pickers';
 import { DateValidationError } from '@mui/x-date-pickers/internals';
 import type { Dayjs } from 'dayjs';
 import React from 'react';
 
+import { useLayout } from '../../../../helpers';
 import { TextFieldProps } from '../../textfield/textfield';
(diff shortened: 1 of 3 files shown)


kasparsuvi1:
#52

libs/react-components/community/src/components/form/pickers/datepicker/datepicker.tsx
```diff
@@ -135,6 +136,10 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     onClose?.();
   };
 
+  // Mui doesn't pick correctly which component to render, thus it was unusable when zooming in on bigger screens
+  const isMobileLayout = useLayout(['mobile']);
```
HarlesPilter, line 140: I get the need behind this change, but I would implement it through `desktopModeMediaQuery` prop https://mui.com/x/api/date-pickers/date-picker/ and possibly add a prop on our end so we could override it on the application level if needed.

Also wondering why MUI team decided this to be based on a touch device and not screen size...

HarlesPilter (changes requested):

libs/react-components/community/src/components/form/pickers/datepicker/datepicker.tsx
kristiina-richards: Missed that prop. Changed to use that

HarlesPilter (approved):

tobre6:
Decided that this will be close

Issue #53, opened 2024-05-23
Layout: horizontal scroll at 400% zoom with 1280px width


kristiina-richards:
### Description of the problem

When using it on 1280px width then 400% zoom has a horizontal scroll. horizontal scroll is fine with tables but not on the whole page at that zoom
**Requirement:** [link](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)

HarlesPilter:
1. Fixed in https://github.com/TEHIK-EE/tedi-design-system/pull/61
2. Will not change currently

PR #61, opened 2024-05-28
fix(layout): fix role for sidenav and add break-all to footer links

kristiina-richards:
https://kristiina-richards.github.io/tedi-design-system/bug/layout-wcag-fixes/?path=/story/community-components-layout--default

### DIFF
--- a/libs/react-components/community/src/components/layout/footer/footer.tsx
+++ b/libs/react-components/community/src/components/layout/footer/footer.tsx
@@ -98,7 +98,7 @@ const FooterCategory = (props: FooterCategory): JSX.Element => {
           <VerticalSpacing element="ul" size={0.5} className={cn(styles['footer__category-list'])}>
             {elements?.map((item, index) => (
               <li key={index}>
-                <Text color="inverted" element="span">
+                <Text color="inverted" element="span" modifiers="break-all">
                   {item}
                 </Text>
               </li>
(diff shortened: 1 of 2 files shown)


kasparsuvi1:
#53

kasparsuvi1:
Navigation role is used on wrong element/style: https://developer.mozilla.org/en-
US/docs/Web/Accessibility/ARIA/Roles/navigation_role

Im not really 100% sure that navigation usage is more correct than menu and menuitem's with link. Terviseportaal was tested for WCAG rules also with real WCAG users and it didn't come to be a problem.

We could add aria-role="navigation" to nav element and remove menu and menuitem roles, so the anchor aria-role is not overwritten. But this needs to be tested, anyway it seems to be more correct way.
W3 examples are done the same way: https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element

When removing menu-item from links we should also check that android and ios screen readers work correctly: https://github.com/TEHIK-EE/tedi-design-system/commit/be5b2c4988a9ba16f50554d3c44a26c49b217662

kristiina-richards:
> Navigation role is used on wrong element/style: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role
> 
> Im not really 100% sure that navigation usage is more correct than menu and menuitem's with link. Terviseportaal was tested for WCAG rules also with real WCAG users and it didn't come to be a problem.
> 
> We could add aria-role="navigation" to nav element and remove menu and menuitem roles, so the anchor aria-role is not overwritten. But this needs to be tested, anyway it seems to be more correct way. W3 examples are done the same way: https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element
> 
> When removing menu-item from links we should also check that android and ios screen readers work correctly: [be5b2c4](https://github.com/TEHIK-EE/tedi-design-system/commit/be5b2c4988a9ba16f50554d3c44a26c49b217662)

I tested and the menu items aren't read as link in that case. But I am happy to take the changes for issue nr 2 back and leave that one still up

HarlesPilter (approved):

PR #62, opened 2024-05-29
feat(colors): TEDI-242 - Added TEDI colors docs

airikej:


tobre6 (approved):

libs/react-components/shared/docs/colors/color-tokens.tsx
```diff
@@ -0,0 +1,55 @@
...
+      }
+      acc[baseName].push(color);
+    });
+    return acc;
+  }, {});
+
+  return (
+    <>
+      {Object.entries(groupedColors).map(([groupName, colors]: [string, Color[]]) => (
+        <div className="color__example__wrapper" key={groupName}>
```
kasparsuvi1, line 32: Why we use custom css for our storybook stories? Could we use the same components, so its would be good example for applications and developers looking storybook?

airikej: I agree, but right now we don't have enough TEDI components to build them and this "custom" style matches the stories in Figma at the moment

kasparsuvi1: why not use community components?

tobre6: We shouldn't have circular dependencies. Nor build dependencies and even in the logical level.
Because TEDI doesn't have control over community components and they can change organically and break TEDI part of the library and storybook.

I agree the thinking that we should reuse. If the components that can be used (utilities, layout, grid components) to build stories are reviewed, tested and moved to TEDI, these components should be revisited and refactored.

tobre6: Also we should think about the roadmap so that these low-level (utility) components should be before.

kasparsuvi1: Don't agree with the breaking part - When component is updated in community side, it also should not break anything. TEDI has control over them (Tedi maintains community components - they are in TEDI repository, TEDI team also can say that some changes are not okay - in code reviews/pull-requests)

For circular dependencies, when community components are used to create layout for TEDI stories it's not really `@tehik-ee/tedi-design-system/tedi` dependencie, because stories doesn't affect the final build of the component-library.

kasparsuvi1: And also applications will start using tedi and community components togheter, thats the goal - We want to use some TEDI components as soon as possible.
So storybook would be the good first place to test out how they compliment each other  and work side-by-side.

tobre6: I have asked this question at TEHIK and let's see what will be the guidelines. But as for now we can only continue what we have agreed and this is that TEDI team doesn't maintain community components. Only watches PRs and doesn't interfere, only if we can give hints.

tobre6 (commented):
TEDI logo is not used anywhere.

PR #72, opened 2024-05-30
[FileUpload]: File rejection helper

MVaik:
Helper text that notifies the user of file rejection causes.


libs/react-components/community/src/components/form/file-upload/file-upload.tsx
```diff
@@ -138,12 +143,22 @@ export const FileUpload = (props: FileUploadProps): JSX.Element => {
   const onFileChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
     if (e.target.files) {
       const invalidSizeFiles: File[] = [];
+      const invalidExtensionFiles: File[] = [];
```
MVaik, line 146: Duplicate logic emerging, perhaps a better idea to combine them into an onReject/onRejected callback and add a type?

MVaik: Proposed alternative: 
```js const rejectedFiles: FileRejected[] = [];
      const uploadedFiles = [...Array.from(e.target.files)]
        .filter((file) => {
          if (validFileType(file)) return true;

          rejectedFiles.push({type: 'extension', file})
          return false;
        })
        .filter((file) => {
          if (!maxSize || file.size <= maxSize * 1024 ** 2) return true;

          rejectedFiles.push({type: 'size', file})
          return false;
        });```

Comment (kasparsuvi1) [reply libs/react-components/community/src/components/form/file-upload/file-upload.tsx:146]: Do application need to to get fallback on failed files? Or maybe its enought when we show error helper from inside FileUpload component? So every application doenst have to filter throught the error types, and combine the error text. 

Because i cant see any more usecases for rejectedFiles than showing error why they were rejected.

Comment (MVaik) [reply libs/react-components/community/src/components/form/file-upload/file-upload.tsx:146]: Showing the helper instead would likely be fine, could show the error if no helper prop was passed in or override it until another upload is attempted.

Comment (MVaik) [reply libs/react-components/community/src/components/form/file-upload/file-upload.tsx]: Reworked the logic to provide a helper instead of callbacks.

File: libs/react-components/community/src/components/form/file-upload/file-upload.tsx [hunk 2]
```diff
@@ -1,4 +1,5 @@
 import cn from 'classnames';
+import { join } from 'path';
```
kasparsuvi1, line 2: Jäänud vale import

kasparsuvi1 (approved):

Issue #102, opened 2024-07-26
[AccordionItemHeader]: WCAG issue


JoelJKoel:
### Description of the problem
Interactive controls must not be nested
Requirement: [link](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)

![image](https://github.com/user-attachments/assets/183d9461-eef5-4978-9dfb-7d7644054fe3)

PR #105, opened 2024-07-26
fix(accordionitemheader): #102 - Swap ToggleOpen Button for an Icon

JoelJKoel:
Swapped the ToggleOpen Button which was used in AccordionItemHeader for an Icon which looks and handles the same getting rid of the nested interactions. 

image


airikej:
@kasparsuvi1 would like second opinion

kasparsuvi1:
Is it needed because AccordionHeader has `role="button"`  and includes button (ToggleOpen) which is invalid html?
Maybe we could use ToggleOpen as not button in that case. Because currently the openText/closeText/isHovered etc funcionality is removed (also props arent). And the ToggleOpen functionality is duplicated to two places.
In summary i can see the problem, but not sure if this is the way to go. And if  we go with this approach, then the PR needs some improvement.

JoelJKoel:
Jah, vahepeal sattusid need tõesti kokku. Aga nagu arutasime peaks ilmselt teisiti lähenema sellele probleemile. Kas proovin ise ToggleOpen komponenti ringi teha et see poleks button või saate te ise mingi hetk sellele peale vaadata?

airikej:
Ma ütleks nii, et kui sellele on vajadus, siis proovi lahendus leida praegu - hetkel raske öelda, millal Accordion Tedi-sse tuleb, sest hetkel pole sellele disainigi.

PR #153, opened 2024-08-26
Feat/tedi 358 component vertical spacing

airikej:


tobre6 (approved):

libs/react-components/tedi/src/components/vertical-spacing/vertical-spacing.spec.tsx
```diff
@@ -0,0 +1,63 @@
+import { render } from '@testing-library/react';
```
kasparsuvi1, line 1: Siin testide hulgas peakid olema ka verticalSpacingItem testid

PR #157, opened 2024-08-29
Fix/tedi 370 heading with icon improvements

airikej:


tobre6 (approved):

airikej:
While I agree with you, the design was dictating a bit giving me only one example. I can bring this topic up with the designers and we can decide how to move on with it further. So your suggestion would be, so that HeadingWithIcon props extend all of HeadingProps and IconProps as well?

airikej:
Discussed this topic with designer and will allow all HeadingProps and IconProps for development

PR #171, opened 2024-09-05
 feat(toggleOpen): #102 - Change ToggleOpen to not be a Button 

JoelJKoel:
To solve the issue #102. 

The problem with these changes is that old `ToggleOpen` supported all `Button` props, with this change only some of them are implemented. Should more optional props be supported and is there a better way to add them?

Solves the accessibility problem and should look and handle the same way in given examples.
image


kasparsuvi1:
I dont think this is the way to go:
1. We are duplicating button styles
2. When ToggleOpen is wanted to use as button (main use-case), it cannot be used liked that.

For example we could use directly `ButtonContent` element, and pass `as` element as div or button depending on the useCase.
Or use  `Button` component inside ToggleOpen and add `as` property to Toggleopen, that can be changed to `div` needed. Im not 100% sure if with the second approach we should also override type="button" from WCAG point of view or is it acceptable and valid HTML.

JoelJKoel:
Thanks for the feedback! Seems like adding an `as` prop to render the button as a `div` when needed was indeed enough to fix the WCAG accessibility issue.

image

libs/react-components/community/src/components/toggle-open/toggle-open.tsx
```diff
@@ -23,20 +23,35 @@ export interface ToggleOpenProps extends Omit<ButtonProps, 'children' | 'iconRig
    * Overidden from ButtonProps, because we dont support string as icon name here
    */
   iconRight?: Partial<IconProps>;
+  /**
+   * If the element is a button or a div
+   * @default button
+   */
+  as?: 'button' | 'div';
```
kasparsuvi1, line 30: PolymorphicComponentPropWithRef types should probaly used here, thought nothing is broken with it.
Poly props are not very straightforward to use, but Sidenav for example has example of it.
@airikej is it ok?

kasparsuvi1 (approved):

libs/react-components/community/src/components/toggle-open/toggle-open.tsx
airikej: Yes I think its fine for now, ready to merge!

airikej (approved):

kasparsuvi1:
@JoelJKoel had to revert after merge, because the build did not complete. #231

PR #173, opened 2024-09-05
fix(general): Modify commitlint rules to include GitHub issue key (#172)

airikej:


kasparsuvi1:
`git commit -m "feat(anything): #173"`

Can we also force that only issue number is not enought for summary?

kasparsuvi1:
image

airikej:
I didn't find a great rule to fix that, but it shouldn't allow to commit with just an issue key anymore

kasparsuvi1 (approved):

Issue #70, opened 2024-05-30
Stepper: Zoom levels affect current step styling


HarlesPilter:
On some zoom levels there might be primary background bleed around the arrow. Probably happens due to browser rounding the decimal px values (39.90px etc). 

Video is done on Chrome Windows
https://github.com/TEHIK-EE/tedi-design-system/assets/16189897/cdbaa7b6-e88a-478f-b6a9-d47406f407fe

PR #186, opened 2024-09-23
fix(stepper): #70 - Background bleed on different zoom levels, active arrow scaling

kaarelrohu:
Due to my unfortunate choice of branch name, I had to close the last pull request. (#184 )

Here is the link to the storybook: https://kaarelrohu.github.io/tedi-design-system/fix/70-stepper-background/?path=/docs/community-stepper--docs


HarlesPilter:
Still have some issues on certain zoom levels 
image
Maybe try to achieve the arrow by creating a absolute positioned square that is under the button and adding margin-right: 1.25rem to the button. Something like this 
image

HarlesPilter (approved):
looks good to me

Issue #181, opened 2024-09-16
FormLabel


airikej:
https://github.com/TEHIK-EE/tedi-design-system/issues/142 

Mostly unchanged from Community's FormLabel

| Name | Required/Optional | Description | Type | Possible values | Default |
|----------|----------|----------|----------|----------|----------|
label | Required | Label value | `string` | | |
id | Required | ID of element label is for | `string` | | |
className | Optional | Additional class name | `string` | | |
required | Optional | Renders required symbol | `boolean` | | `false` |
requiredLabel | Optional | Label shown when input is required | `string` | | |
hideLabel | Optional | Whether to hide the label visually. Use keep-space to save label space | `boolean \| 'keep-space'` | | `false` |
renderWithoutLabel | Optional | Render without label HTML element - used when Form component uses label inside already | `boolean` | | `false` |

**Need clarifying:**

- `size` prop - In Community's FormLabel there is a size prop, that has values: `'small' | 'default' | 'large'` and `size` prop is only used inside TextField and NumberField components. Is it realistically needed? Where would you use large sized FormLabel?
- Would you benefit from `isBold` prop? It would render the label as bold.

kasparsuvi1:
`size` - smaller inputs would have smaller labels. So when u use select or textfield that is size="small". Then the label would be smaller in size also (atleast with previous design)

`isBold` - It's question to design, should there be possibility to use bold label on form elements? I dont think so, atleast currently i don't know use-case where it should be needed. And it doesn't hold applications with one style imo.

HarlesPilter:
We use large TextField in Terviseportaal digital-registry flow
![image](https://github.com/user-attachments/assets/819c2fdb-3a58-4396-9d0d-3c638f06e67c)

airikej:
@karolin-kivisikk It appears all sizes are in use, including large.

karolin-kivisikk:
> @karolin-kivisikk It appears all sizes are in use, including large.

* Added small label size to Figma
* Added Large input size under [Search](https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=4620-82860&t=cZyksISdQXcCpHW0-4) component

PR #214, opened 2024-10-07
feat(form-label): Component Label and FormLabel #181

airikej:


mart-sessman (approved):

libs/react-components/tedi/src/components/form/form-label/form-label.tsx
```diff
@@ -0,0 +1,70 @@
...
+  renderWithoutLabel?: boolean;
+  /**
+   * Specifies the size of the label text.
+   * Options include 'small' for a smaller label size or 'default' for the standard size.
+   * @default default
+   */
+  size?: 'small' | 'default';
+}
+
+export const FormLabel = (props: FormLabelProps): JSX.Element => {
```
kasparsuvi1, line 44: Is there some problems why we cant use this component inside community form components?

airikej: Do you mean replace the old one and remove the Community's FormLabel?

kasparsuvi1: Yes, not sure why we need duplicated components.

airikej: Yes agreed, just forgot to do that, will make a task and replace them and make sure nothing will be broken

PR #229, opened 2024-10-15
feat(form-label): Replace FormLabel in Community components with TEDI-…

airikej:
…Ready FormLabel #227


kasparsuvi1:
Can't really publish it under "Bug fix" . This PR removes `requiredLabel` from all form components. Change it atleast to feature or Breaking change as it should be.

kasparsuvi1 (approved):

Issue #318, opened 2024-11-22
[Icon]: New Tedi Icon component isn't wrapped with forwardRef


kasparsuvi1:
## Problem
Community Icon was wrapped with forwardRef, so it could be used directly as TooltipTrigger and other like components children, that uses it's children ref for something. TooltipTrigger for example gives oppurtunity to open tooltip on hover/click etc.

## Solution
Tedi Icon should be wrapped same way as Community component was. To allow usages like that:
```
                    <TooltipProvider placement="top">
                      <TooltipTrigger>
                        <Icon name="info" />
                      </TooltipTrigger>
                      <Tooltip>
                        Tooltip content
                      </Tooltip>
                    </TooltipProvider>
```

PR #321, opened 2024-11-25
fix(icon): Added forwardRef possibility to Icon #318

airikej:


kasparsuvi1:
Doesn't work - React.forwardRef needs to be used. For testing please use Tedi Icon inside tooltip stories.

kasparsuvi1 (approved):

PR #322, opened 2024-11-25
fix(link): Allow link wrapping into multiple lines #316

airikej:


HarlesPilter:
Icon doesn't align with text 
image

Issue #320, opened 2024-11-22
[Alert]: Content only alignment


HarlesPilter:
When there is no title, there is a noticable shift in vertical alignment

![image](https://github.com/user-attachments/assets/9aa47fbd-f104-4e6e-912b-1a40c852dda2)

Seems to be that VerticalSpacing adds margin bottom to all children, but last element shouldn't need one

![image](https://github.com/user-attachments/assets/0cf4f29c-8093-4622-9535-b348439731ed)

HarlesPilter:
Win + 131.0.6778.86 Chrome. Still there. Best visible with this story: https://github.tehik.ee/tedi-design-system/rc/?path=/docs/tedi-ready-components-notifications-alert--docs#alert%20colors

airikej:
Thank you, will check it out

PR #324, opened 2024-11-25
fix(alert): Fix content only alignment #320

airikej:


HarlesPilter (changes requested):
I think the fix should be that when no title is present then don't render
`<div className="tedi-alert__content-wrapper">{title && children}</div>`. Right now we sould render an empty div.

libs/react-components/src/tedi/components/alert/alert.tsx
```diff
@@ -130,7 +130,7 @@ export const Alert = (props: AlertProps): JSX.Element | null => {
             </Col>
           )}
         </Row>
-        <div className="tedi-alert__content-wrapper">{title && children}</div>
+        {title && children && <div className="tedi-alert__content-wrapper">{title && children}</div>}
```
HarlesPilter, line 133: No need to double check title inside the div

HarlesPilter (approved):

Issue #347, opened 2024-12-05
[Link]: Icon not vertically center


airikej:
**Problem:**

Current Link Icon doesn't seem vertically center. It appears a bit lower compared to text

PR #349, opened 2024-12-05
fix(link): Icon not vertically center #347

airikej:


libs/react-components/src/tedi/components/buttons/button-content/button-content.tsx
```diff
@@ -138,7 +138,15 @@ const InternalButtonContent = forwardRef(
       const iconBEM = cn(styles['tedi-btn__icon'], styles[`tedi-btn__icon--${location}`], {
         [styles['tedi-btn__spinner']]: isLoading,
       });
-      const defaultIconProps: Partial<IconProps> = { size: 18, className: iconBEM };
+
+      const isAnchor = Component === 'a';
```
HarlesPilter, line 142: Kui rakenduses kasutatakse React routeri või Next routeri, siis Component on Link

libs/react-components/src/tedi/components/buttons/button-content/button-content.tsx
```diff
@@ -138,7 +138,15 @@ const InternalButtonContent = forwardRef(
       const iconBEM = cn(styles['tedi-btn__icon'], styles[`tedi-btn__icon--${location}`], {
         [styles['tedi-btn__spinner']]: isLoading,
       });
-      const defaultIconProps: Partial<IconProps> = { size: 18, className: iconBEM };
+
+      const isAnchor = as === 'a' || !!rest.href;
```
HarlesPilter, line 142: Not all links will have a href. For example React routers Link uses `to` prop. We should use something else to make the distinction. When using a button, that will visually have link styles the icon should be inline as well. So maybe use visualType === 'link' ?

airikej: Yes that's a good idea, no clue why I overthought it

HarlesPilter (approved):

Issue #234, opened 2024-10-17
Text field


airikej:
https://github.com/TEHIK-EE/tedi-design-system/issues/119 

[Figma](https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=3486-37651&m=dev)

Dependencies:
- FormLabel
- Button
- Icon
- Separator

No API changes but would suggest this:
- Merge `onChange` and `onChangeEvent`: This can simplify the API by having a single prop that supports both use cases.
- Add unit tests

kasparsuvi1:
> Merge onChange and onChangeEvent: This can simplify the API by having a single prop that supports both use cases.
Most common cases use onChange, because there is nothing else but value needed from the event.
But there are some usecases where event is needed, for that purpose i think both should be available.

airikej:
@kasparsuvi1 @HarlesPilter is TextField with clearing button used anywhere else than just as search input? The designers want to break them into 2 components, so TextField wouldn't be having either large variant and also clearable button - they would be exclusive to Search component later. What do you think about that?

PR #352, opened 2024-12-06
feat(textfield): TEDI-Ready TextField development #234

airikej:


kasparsuvi1:
#234

libs/react-components/src/tedi/components/form/textfield/textfield.module.scss
```diff
@@ -0,0 +1,205 @@
...
+  cursor: initial;
+}
+
+.tedi-textfield__icon {
+  vertical-align: text-top;
+}
+
+/* Small */
+.tedi-textfield--small {
+  & .tedi-textfield__input {
```
mart-sessman, line 161: Üleliigne tühik?
All pool ka, näiteks rida 183, 188.

mart-sessman (changes requested):

libs/react-components/src/tedi/components/form/textfield/textfield.module.scss
airikej: This is not a bug and works as if directly nesting child class inside. But i removed the & for clarity

mart-sessman (approved):

libs/react-components/src/community/components/form/select/select.module.scss
```diff
@@ -77,8 +79,9 @@ div .select__menu {
 
 .select--small {
   & .select__control {
-    min-height: 1.875rem;
-    font-size: 0.75rem;
+    height: $input-height-small;
+    min-height: $input-height-small;
+    font-size: 0.85rem;
```
kasparsuvi1, line 84: This not valid size - 16px * 0.85 = 15,3px.

PR #357, opened 2024-12-09
fix(modal): Modal open focus fix #343

### LINKED ISSUE #343: [Modal]: Modal is not focused when opened
When navigating with keyboard the modal does not receive focus. 
After focusing inside the modal, then focus is not trapped as well and leaves modal after last focusable element
Repro can be done in all modal stories https://github.tehik.ee/tedi-design-system/rc/?path=/story/community-modal--controlled-outside

This was definitely working at some point https://github.tehik.ee/tedi-design-system/gitlab-master/?path=/story/components-modal--default

airikej:


HarlesPilter:
Any idea why this changes fixes it and why it broke?
Also I can't find a link for the deployed storybook

airikej:
> Any idea why this changes fixes it and why it broke? Also I can't find a link for the deployed storybook

https://github.tehik.ee/tedi-design-system/fix/343-modal-modal-is-not-focused-when-opened/?path=/docs/community-modal--docs
No idea why it wasn't deployed before. The code i removed, I've added earlier to combat the ClosingButton focus "problem" - the ClosingButton was in focused state the moment modal was opened, but that ref apparently broke the modal focus itself

HarlesPilter (approved):

Issue #344, opened 2024-12-03
[Icon]: className is not passed to wrapper


HarlesPilter:
When className is defined on TEDI Icon then it is passed to the icon span and not the wrapper div
The className should be passed to the wrapper div instead of if that is a breaking change then add a separate prop for passing className to wrapper div

PR #359, opened 2024-12-10
fix(icon): Apply className to wrapper when background is set #344

airikej:


HarlesPilter (commented):
Also update the className props comment with info that sometimes it is applied to the icon and sometimes to the wrapper

HarlesPilter (approved):

Issue #261, opened 2024-10-28
Card


airikej:
https://github.com/TEHIK-EE/tedi-design-system/issues/116 

[Figma](https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=163-19532&m=dev)

Wouldn't be changing anything in API, except cleanup for deprecated props and matching naming conventions with Figma design.
Card already has BreakpointSupport, so no need to add that.
Currently only unit tests are missing.

Any other suggestions are welcome

kasparsuvi1:
Also new TEDI design shows card's with background images. Currently its not supported.  And custom solution is needed by applications.
Can we add backgroundImage property, that allows to pass base64 string image and also placement properties of the image? So the custom css is not needed by applications.

PR #362, opened 2024-12-13
feat(card): TEDI-Ready Card component development #261

airikej:


libs/react-components/src/tedi/components/cards/card/card-stories-templates.tsx
```diff
@@ -0,0 +1,546 @@
...
+        </Card>
+      </Col>
+    </Row>
+  </VerticalSpacing>
+);
+
+export const CardInfoTemplate: StoryFn<CardStory> = () => (
+  <VerticalSpacing>
+    <Row>
+      <Col>
```
kasparsuvi1, line 341: This Row/Col combo is not needed, VerticalSpacing already gives space between Cards.

libs/react-components/src/tedi/components/cards/card/card-stories-templates.tsx
```diff
@@ -0,0 +1,546 @@
...
+              </Col>
+            </Row>
+          </CardContent>
+        </Card>
+      </Col>
+    </Row>
+  </VerticalSpacing>
+);
+
+export const CardInfoTemplate: StoryFn<CardStory> = () => (
```
kasparsuvi1, line 338: All the repeating code would be nice to be removed and moved into functions, that take needed properties inside.

kasparsuvi1:
Terviseportaal recently added CardContent with background --green-200 from example's it seems like there is no such background. Background template only show 9 colors, from code it seems it allows more, can we demo all of them?
image

Also changing properties for storie doesn't add them to component, show its impossible to test every background/property from prop table.

libs/react-components/src/tedi/components/cards/card/card-content/card-content.tsx
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Background size for the image.
+   */
+  backgroundSize?: React.CSSProperties['backgroundSize'];
+  /**
+   * Background repeat for the image.
+   */
+  backgroundRepeat?: React.CSSProperties['backgroundRepeat'];
+  hasSeparator?: boolean;
+  stretch?: boolean;
```
kasparsuvi1, line 57: What `stretch` does?

libs/react-components/src/tedi/components/cards/card/card-header/card-header.tsx
```diff
@@ -0,0 +1,97 @@
+import cn from 'classnames';
+import React from 'react';
+
+import { BreakpointSupport, useBreakpointProps } from '../../../../helpers';
+import styles from '../card.module.scss';
+import CardContent, { CardContentPadding, CardContentProps } from '../card-content/card-content';
+import { CardContext } from '../card-context';
+import { CardBackground, getPaddingCssVariables } from '../utility';
+
+export type CardHeaderBreakpointProps = {
```
kasparsuvi1, line 10: When it's duplicated props from CardContent and we render inside a CardContent, why we cant reuse the props?

When we want to still export CardHeaderProps separately, we can sue something like:
`export type CardHeaderProps = CardContentProps` (but i dont think it's really necessary)

libs/react-components/src/tedi/components/cards/card/card-stories-templates.tsx
airikej: I replaced the rows/cols with VerticalSpacing and while I agree, it's an ugly long code but since it's in a separate file like (card-stories-templates) then maybe it's useful for someone to copy that template into their project? It's only descriptive file with examples

libs/react-components/src/tedi/components/cards/card/card-content/card-content.tsx
```diff
@@ -100,7 +49,6 @@ export const CardContent = (props: CardContentProps): JSX.Element => {
       data-padding={typeof padding === 'number' ? `${padding}rem` : undefined}
       className={cn(styles['tedi-card__content'], backgroundClass, className, {
         [styles['tedi-card__content--separator']]: hasSeparator,
-        [styles['tedi-card__content--stretch']]: stretch,
       })}
       style={{ ...getPaddingCssVariables(padding), ...backgroundStyle }}
```
kasparsuvi1, line 53: Should the props.style be added here too?

libs/react-components/src/tedi/components/cards/card/card-stories-templates.tsx
kasparsuvi1: Copy from directly code? Because the template is not shown in Storybook currently.
image

To copy the code, the itself doenst have to be written in that way. For example take community/Icon stories -> It has Templates that include functions and passing props, but in storybook the final React code is shown:
image

libs/react-components/src/tedi/components/cards/card/card.tsx
```diff
@@ -0,0 +1,115 @@
...
+  padding?: CardContentPadding;
+  /**
+   * Background color.
+   * @default primary
+   */
+  background?: CardBackground;
+  /**
+   * Background image.
+   */
+  backgroundImage?: string;
```
mart-sessman, line 40: CSSProperties['backgroundImage'];

mart-sessman (approved):

libs/react-components/src/tedi/components/cards/card/card-header/card-header.tsx
```diff
@@ -0,0 +1,14 @@
+import cn from 'classnames';
+
+import styles from '../card.module.scss';
+import { CardContent, CardContentProps } from '../card-content/card-content';
+
+export type CardHeaderProps = CardContentProps;
```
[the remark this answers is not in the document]
airikej: Yep, there is no need, will use CardContentProps instead

Issue #339, opened 2024-12-02
Hide on scroll


mart-sessman:
Props:
| Name | Required/Optional | Description | Type | Default |
|----------|----------|----------|----------|----------|
children | Required | Any content to be rendered within the component | `React.ReactNode` | |
className | Optional | Additional class name | `string` | |
enabled | Optional | Is scrolling behaviour enabled | `boolean` | `true` |
visibility | Optional | Hide/show when scrolling past scrollDistance | `'hide' \| 'show'` | `'hide'` |
toggleVisibility | Optional | Toggle visibility when scrolling to opposite direction | `boolean` | `false` |
scrollDistance | Optional | Distance in px to hide/show | `number` | `100` |
scrollDirection | Optional | Direction used to calculate `scrollDistance` | `'up' \| 'down'` | `'down'` |
scrollContainer | Optional | Detect scroll on this parent element | `HTMLElement` | `document.documentElement` |
animationDirection | Optional | Animate children to this direction | `'left' \| 'right' \| 'down' \| 'up' \| 'center'` | `'center'` |

PR #363, opened 2024-12-13
feat(hide-on-scroll): add component #339

mart-sessman:


kasparsuvi1:
HideOnScroll should be separate PR and also delete commmunity version of it?

mart-sessman:
Yes, Floating Button should not be in this PR. Fixing this.

Also renaming this HideOnScroll component to ScrollVisibility, because it does more than just hiding, like an example: showing content when scrolling past scroll distance. But this will be done in separate task.

Since this component's name won't be "HideOnScroll", the community version will stay, but will be deprecated and new component linked to it.

mart-sessman:
You are right, keeping 2 components with the same functionality is not reasonable.
Removed old component and old hook.
Also there is one BREAKING CHANGE.

Old component behaved this way:
When scrolled past "scrollDistance" prop and then scrolled back up 10px, the content would show again.
This behaviour is still present, but disabled by default.
Now for the same behaviour, "toggleVisibility" prop must be set to true.

libs/react-components/src/tedi/components/hide-on-scroll/hide-on-scroll.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+const CONTAINER_HEIGHT = 300;
+
+/**
+ * <a href="https://tedi.tehik.ee/1ee8444b7/p/00ab5a-scrollvisibility-dev" target="_BLANK">Zeroheight ↗</a>
+ **/
+
+const meta: Meta<typeof HideOnScroll> = {
+  component: HideOnScroll,
+  title: 'Tedi-ready/Components/HideOnScroll/HideOnScroll',
```
airikej, line 17: This component could be placed inside Helpers category, at least for now

airikej (approved):

Issue #364, opened 2024-12-13
Truncate


mart-sessman:
https://github.com/TEHIK-EE/tedi-design-system/issues/111

Sõltuvus:
- Button

mart-sessman:
Props:
| Name | Required/Optional | Description | Type | Default |
|----------|----------|----------|----------|----------|
children | Required | The text content to truncate | `string` | |
className | Optional | Additional class name | `string` | |
length | Optional | Maximum number of characters to display, takes priority over 'lines' prop. If not provided use line-based truncation | `number` | |
lines | Optional | Maximum number of lines to display. | `number` | `1` |
ellipsis | Optional | Custom content to display at the end of truncated text | `ReactNode` | `'...'` |
expandable | Optional | Whether the truncated text should be expandable | `boolean` | `false` |
button | Optional | Override default button props | `Partial<ButtonProps>` | |

PR #370, opened 2024-12-16
feat(truncate): TEDI-Ready Truncate development #364

mart-sessman:


libs/react-components/src/tedi/components/truncate/truncate.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ellipsis = '...',
+    expandable = true,
+    button,
+  } = getCurrentBreakpointProps<TruncateProps>(props);
+  const { getLabel } = useLabels();
+
+  const [isTruncated, setIsTruncated] = useState(true);
+  const truncatedText = useMemo(() => {
+    let slicedText = children.slice(0, maxLength);
+    slicedText = slicedText.replace(/[^a-zA-Z0-9]+$/, '');
```
airikej, line 58: I wonder if this could be replaced with something like this `slicedText = slicedText.trimEnd();`
have you tested it?

libs/react-components/src/tedi/components/truncate/truncate.tsx
```diff
@@ -0,0 +1,82 @@
...
+      {isTruncated ? truncatedText : children}
+      {expandable && (
+        <Button
+          style={{ display: isTruncated ? undefined : 'block' }}
+          visualType="link"
+          {...button}
+          onClick={(e) => {
+            button?.onClick?.(e, isTruncated);
+            setIsTruncated((prev) => !prev);
+          }}
```
airikej, line 73: Wcag would benefit from `aria-expanded={!isTruncated}` as well

libs/react-components/src/tedi/components/truncate/truncate.tsx
mart-sessman: Yes, works like this also. Replacing it. Shorter code.

airikej (approved):

Issue #376, opened 2024-12-18
[Check, Radio]: Alignment of tooltip is not centered


kasparsuvi1:
## Problem
![image](https://github.com/user-attachments/assets/3ae44c2e-1904-4131-b297-8a7afc5f87b0)

## Solution
Check and Radio tooltip alignment should be center of the label. Currently every inline icon doesnt align when used directly inside Text. (Community or Tedi, something has changed.)

airikej:
I checked the old gitlab branch, which has pre tedi stage still to compare things and frankly, they have never been vertically center for radio/check:
https://github.tehik.ee/tedi-design-system/gitlab-master/?path=/docs/components-form-radio--docs
https://github.tehik.ee/tedi-design-system/gitlab-master/?path=/docs/components-form-check--docs

but i've made the fix, seen here:
https://github.tehik.ee/tedi-design-system/fix/376-check-radio-alignment-of-tooltip-is-not-centered/?path=/docs/community-form-check--docs
https://github.tehik.ee/tedi-design-system/fix/376-check-radio-alignment-of-tooltip-is-not-centered/?path=/docs/community-form-radio--docs
https://github.com/TEHIK-EE/tedi-design-system/pull/386

kasparsuvi1:
## Problem
There seems to be multiple issues how we use icons in Design and development. 

Currently `Icon` component doesn't allow to use inline Material symbols as its recommended by Material team next to any text.

Main points where current Tedi Icon doesn't meets requirements or guidelines of Material symbols:
1. Use the same size for your Material Symbols and text
2. Use the same optical weight for your symbol and text
3. Shift down the baseline of symbols to approximately 11.5% of the text size
(Should be confirmed with Tedi designers also, because it's not specified in Tedi Figma)

[See reference](https://m3.material.io/styles/icons/applying-icons#f9db4adc-ca78-473f-85eb-a351b73c39ac)

## Possible solutions
1. Inline icons should be `font-size: inherit`, this way there is no possibility to use icon inline of the text in the wrong size.
 1.1. We should allow every possible font-size of Tedi Text and Heading to be also Icon size. (When use inside Row/Col)
 1.2. I think that Icons with background shouldn't be allowed to use as inline icons. Because from design point of view they are 
       never meant to be used like that (should also remove examples of that)
2.

kasparsuvi1:
## Problem
There seems to be multiple issues how we use icons in Design and development.

[...]

PR #386, opened 2024-12-22
fix(check,radio): Alignment of tooltip is not centered #376

airikej:


libs/react-components/src/tedi/components/icon/icon.module.scss
```diff
@@ -83,6 +82,11 @@ $icon-sizes: (
     }
   }
 
+  &--size-inherit {
+    font-size: inherit;
+    vertical-align: -11.5%;
+  }
+
```
kasparsuvi1, line 89: This should be added to the `tedi-icon--inline` class not with separate property. All inline Icons should be the same size as font-size. As material-symbols docs say.

libs/react-components/src/tedi/components/icon/icon.tsx
```diff
@@ -3,7 +3,7 @@ import { forwardRef } from 'react';
 
 import styles from './icon.module.scss';
 
-export type IconSize = 8 | 12 | 16 | 18 | 24 | 36 | 48;
+export type IconSize = 8 | 12 | 16 | 18 | 24 | 36 | 48 | 'inherit';
```
kasparsuvi1, line 6: There are still missing icon sizes to cover all possible text-sizes (in mobile, tablet and desktop).
Use case: Icon is rendereded not inline but for example in Row/Col next to a text. It should be possible to match the icon size with text size. Currently it is not.

libs/react-components/src/tedi/components/icon/icon.tsx
kasparsuvi1, line 1: From the issue comment: 
```
1.2. I think that Icons with background shouldn't be allowed to use as inline icons. Because from design point of view they are
never meant to be used like that (should also remove examples of that and add conditional types)
```

Currently it's still possible to use background icons as inline we should typesafe that.

kasparsuvi1 (changes requested):

kasparsuvi1 (approved):

Issue #379, opened 2024-12-18
HashTrigger


mart-sessman:

mart-sessman:
This component's functionality is like link/anchor component's.
Changing the URL and moving to correct part in the page.
Move this to Link component?

mart-sessman:
Currently HashTrigger has 2 props: "id" and "scrollOnMatch".
When moving this to Link componenet, I would suggest to add only one prop: "hash"

If "hash" is defined, then Link component gets HashTrigger functionality:

Looking for hash in the url
Looking for node which "id" is equal to hash.
If hash is not defined, then it will work as normal Link.

mart-sessman:
Currently HashTrigger component does not work on it's own if child node does not have "id" attribute.
HashTrigger component has "id" prop, but it is only used for finding an element on the page and scrolling to it.
If we don't add "id" attribute explicitly to child node, then it does not work.

I suggest forwarding "id" prop from HashTrigger to child element. If children is string or some other node, which is not HTML element, then we add wrapper div to children and forward "id" prop to it.

But this also means that we have to change TabsItem component, remove HashTrigger from it and put it into TabsNavItem component, wrapping Anchor component there.

mart-sessman:
@kasparsuvi1 maybe you have some takes on this, since you have worked with these components before?

PR #396, opened 2025-01-13
feat(hash-trigger): TEDI-Ready component development #379

mart-sessman:
HashTrigger previously did not work when child element inside did not have "id" prop.
Improved component, so it would pass the "id" to child element.
Since TabsItem and AccordionItem both used this component, they need changes too.

Previously TabsItem used HashTrigger component, but it was used for for detecting the scroll.
The id was actually added to TabsNavItem component.
Since now the HashTrigger component also adds "id" to child element, I moved it from TabsItem to TabsItemNav component. Everything should work the same as before.

Same goes for AccordionItem and AccordionItemHeader.


mart-sessman:
@HarlesPilter thoughts?

libs/react-components/src/community/components/tabs/tabs-item/tabs-item.tsx
```diff
@@ -26,33 +25,25 @@ export interface TabsItemProps extends BreakpointSupport<TabsItemBreakpointProps
...
+  const { currentTab } = React.useContext(TabsContext);
   const isCurrent = id === currentTab;
 
+  if (!isCurrent) {
+    return null;
+  }
+
   return (
-    <HashTrigger
-      data-name="tabs-item"
```
HarlesPilter, line 40: Removing this might potentially cause a breaking change, since some tests on the app level might use this to target content. By adding this to the div it should be okay.

libs/react-components/src/tedi/components/hash-trigger/hash-trigger.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    window.addEventListener('hashchange', handleHashChange);
+
+    return () => {
+      window.removeEventListener('hashchange', handleHashChange);
+    };
+  }, [id, onMatch, scrollOnMatch]);
+
+  return isValidElement(children) ? cloneElement(children, { id } as { id?: string }) : <div id={id}>{children}</div>;
```
HarlesPilter, line 75: Hmm. But what if the child element does not have an id prop and doesn't pass props down the html with ...rest etc. For example if the HashTrigger is used around a custom component created in the app. Because we can't really check if the child component supports id prop or what it does with it, then it should be documented somewhere that the child has to accept the id prop and inject it into the DOM as well

mart-sessman: I added description to id property.

libs/react-components/src/community/components/tabs/tabs-item/tabs-item.tsx
mart-sessman: Added this back, did not want to remove it.

HarlesPilter (approved):

Issue #406, opened 2025-01-15
@tehik-ee/tedi-react package styles are not complete


kasparsuvi1:
## Problem
* general typography styles are missing
* Icon styles are missing

## Examples and information
1. https://github.tehik.ee/tedi-design-system/rc/react/?path=/docs/community-typography-text--docs
2. Try to use Tedi Icon outside of storybook

PR #403, opened 2025-01-14
feat(general): tedi-core, tedi-angular and tedi-react monorepo #330

airikej:

PR #408, opened 2025-01-15
fix(general): @tehik-ee/tedi-react package styles are not complete #406

airikej:
https://github.tehik.ee/tedi-design-system/fix/406-tehik-eetedi-react-package-styles-are-not-complete/react/?path=/docs/community-typography-text--docs

missing typography classes should be included now, but i couldn't reproduce missing TEDI Icon styles, the classes and fonts were included in dist folder and I tested the package in a random react project and icon styles were applying


kasparsuvi1:
Are you sure there is no more mistakes made in #403?
These 3 places were only the ones i accidentally looked into, when these are old bugs somehow merged into this PR im sure there will be more... Also i added some comments to #403, there are changes in eslint, commitlin etc configurations that i don't think should be change .

kasparsuvi1 (approved):

Issue #377, opened 2024-12-18
Affix


mart-sessman:

mart-sessman:
Moving to TEDI-Ready.
Properties stay the same.

PR #411, opened 2025-01-16
feat(affix): TEDI-Ready component development #377

mart-sessman:


libs/react-components/src/tedi/components/affix/affix.tsx
```diff
@@ -12,14 +12,14 @@ export interface AffixProps {
...
-  children?: React.ReactNode;
+  children?: ReactNode;
   /**
    * Additional class.
    */
   className?: string;
   /**
    * Position of Affix.
-   * @default 'sticky'
+   * @default sticky
```
airikej, line 22: did linter remove those single quotes from here?

mart-sessman: No, removed it manually, because in storybook there were double quoutes.

airikej (approved):

Issue #389, opened 2025-01-06
[Separator]: Inline vertical Separator in text


HarlesPilter:
## Problem
Currently `<Separator axis="vertical" />` does not work well when it is inlined in a text
It should be possible to use `<Text>Some random text <Separator axis="vertical" /> that is separated</Text>`
Also size prop does not work when axis="vertical"

## Solution
Add a story with the above example and make sure that separator is visible, and the left/right spacing is also configurable. When the text wraps then the separator height should stay the same as texts line-height

PR #427, opened 2025-01-20
fix(separator): Inline vertical Separator in text #389

airikej:
https://github.tehik.ee/tedi-design-system/fix/389-separator-inline-vertical-separator-in-text/react/?path=/docs/tedi-ready-components-helpers-separator--docs

### DIFF
--- a/libs/react-components/src/tedi/components/separator/separator.module.scss
+++ b/libs/react-components/src/tedi/components/separator/separator.module.scss
@@ -38,6 +38,14 @@ $sizes: (
   &--accent {
     border-color: var(--general-border-accent);
   }
+
+  &--block {
+    display: block;
+  }
+
+  &--inline {
+    display: inline;
+  }
 }
 
 .tedi-separator--dotted,
@@ -45,7 +53,6 @@ $sizes: (
   &::before {
     position: absolute;
     top: 1.25rem;
-    display: block;
     content: '';
     background-color: var(--general-border-primary);
     border-radius: 100%;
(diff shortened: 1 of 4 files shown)


libs/react-components/src/tedi/components/separator/separator.tsx
```diff
@@ -39,6 +39,10 @@ export interface SeparatorSharedProps {
    * @default 1
    */
   thickness?: 1 | 2;
+  /**
+   * Spacing on top and bottom of separator. Only for horizontal axis.
```
HarlesPilter, line 43: If it is supported for both axis then change the comment

libs/react-components/src/tedi/components/separator/separator.module.scss
```diff
@@ -38,14 +38,21 @@ $sizes: (
   &--accent {
     border-color: var(--general-border-accent);
   }
+
+  &--block {
+    display: block;
```
HarlesPilter, line 43: display: block on line 21 should also be removed

libs/react-components/src/tedi/components/separator/separator.tsx
```diff
@@ -39,6 +39,10 @@ export interface SeparatorSharedProps {
    * @default 1
    */
   thickness?: 1 | 2;
+  /**
+   * Spacing on top and bottom of separator.
```
HarlesPilter, line 43: Spacing on top/bottom or left/right based on the axis

HarlesPilter (approved):

Issue #429, opened 2025-01-20
[Typography]: text-normal css class specificity is higher than other modifiers


kasparsuvi1:
## Problem
After #403 the order how the CSS classes are imported changed. Because of that combos like `<Heading element="h4" modifiers={['normal', 'bold']} color="primary">` are not working anymore and it's not possible to achieve semantically correct h4 element with body text properties and some special modifiers. Like bold, different color etc.

## Solution
Classes like `text-normal, h1, h2 ,h3` etc should come before common modifiers like `bold, color-classes, small,  italic` etc to allow usage of more specific modifiers.

## Examples and information
```
<Heading element="h4" modifiers={['normal', 'bold']} color="primary">
 Some heading
</Heading>
```
Should render h4 element with normal font-size and line-height in primary color and bold.
Currently it renders just text-normal.

airikej:
I have one question - should .text-normal even have a color by itself? I figure it could only apply base font size/line-height/weight but color could be applied with color prop on text?

kasparsuvi1:
Possible probally yes, in earlier versions the H1 text was default text-primary/brand color and i assume it was added because of that. Currently i don't think we have any element/modifier that sets color, except color prop itself.

PR #431, opened 2025-01-20
fix(typography): text-normal css class specificity is higher than other modifiers #429

airikej:
Removed duplicate modifiers from community's _typography.scss file as they are already present in tedi-core


kasparsuvi1:
This doesn't fix the same issue for Tedi Text component:
```
<Text element="h4" modifiers={['normal', 'bold']} color="warning"
```
CSS added:
image
Which means that we can't overwrite text-normal with any colors.
We could also benefit from examples that demo that use-case.

Also Stroybook loads double css, maybe it was also side effect from #403? (some rules even triple or more times)
image

kasparsuvi1:
Example in Tedi/Heading , i added real normal-text example to the bottom:
image

Last Heading in code is: 
```
      <Heading element="h1" modifiers={['normal']}>
        H1 element with normal text styles
      </Heading>
```
This should be rendered same as the added body text to the bottom, currently its still with H1 styles.
Is there anything confusing about how this is supposed to work?

kasparsuvi1 (approved):

PR #444, opened 2025-01-23
fix(separator): Dotted small variant size fix #443

### LINKED ISSUE #443: [Separator]: dotted small variant is rendering as large
## Problem
As from design `Vertical Dotted Small` Separator should have dot of 9x9px. Current [examples](https://github.tehik.ee/tedi-design-system/rc/react/?path=/story/tedi-ready-components-helpers-separator--vertical-dotted-small) it is still 15px.

Also we have now variant and dotSize properties. That conflict with each-other
And why do we need dotted only Separator, it doesn't seem like it should be separator then.

## Solution
There should be possibility to use small and large variant of dotted separator.

airikej:


kasparsuvi1 (approved):
I still think the properties are misleading. 
```
<Separator variant="dotted" dotSize="extra-small" />
```
This example code would make sense to render dotted separator line with extra-small dot, but it doesn't.

When someone is just looking at properties it doesn't make sense. And it's not force-type'd so dot-size couldn't be used when variant="dotted" is not used.

It would be better if it would be type'd that way so it can't be used wrongly or should move dot-size into variant. As dot-only-small etc.

Issue #445, opened 2025-01-24
[Heading]: Headings without modifiers don't get correct styles


kasparsuvi1:
## Problem
When using community component as:
```
<Heading>Heading 1</Heading>
```
It renders as <h1>Heading</h1>, which is correct, but i doesn't get correct heading styles. Before it was applied without class directly with h1 element.

Same problem is with Tedi component.

PR #449, opened 2025-01-24
fix(typography): Headings without modifiers don't get correct styles #445

airikej:


kasparsuvi1:
We still should add styles directly to heading elements. Currently we supported <h1>Some header</h1>, we shouldn't brake that i think.

airikej:
It should work now using plain HTML semantic tags for headings as well, I've displayed both examples here:
https://github.tehik.ee/tedi-design-system/fix/445-heading-headings-without-modifiers-dont-get-correct-styles/react/?path=/story/tedi-ready-base-typography-heading--semantic-headings

kasparsuvi1:
Okay, then the issue was different. And this change shouldn't be needed, because the example u did worked before the same? (U can try to revert all the text.tsx changes and the example still works)

Image of style applied to h1, when used: `<Heading>Heading</Heading>` inside application.
image

I didn't exactly say inside #445 what is the issue - sorry. But as i said in the issue the problem was with community component - ad i tried now the same problem is with Tedi component (The fix is only applied to Tedi component currently. Should fix both.)

Real issue is that font-weight is over prioritize from _base.scss file. When components are built together. (In storybook build the sequence is different

kasparsuvi1:
Also please improve the SemanticHeadings example:
* h2 is missing
* h1 example should work as `<Heading>Heading</Heading>`. Without element passed, because h1 is default value.

airikej:
I've improved the stories of both TEDI and Community Headings. 
TEDI Heading: https://github.tehik.ee/tedi-design-system/fix/445-heading-headings-without-modifiers-dont-get-correct-styles/react/?path=/docs/tedi-ready-base-typography-heading--docs
Community Heading: https://github.tehik.ee/tedi-design-system/fix/445-heading-headings-without-modifiers-dont-get-correct-styles/react/?path=/docs/community-typography-heading--docs
Also I understood what the problem was:
1. `<h1>`, `<h2>`, etc.. didnt't take headings styles before (now it does)
2. _base.scss font weight was overriding _typography.scss font weights for headings, it should be fixed now too as I've updated the order in tedi-core and investigated the order or dist/index.css classes

kasparsuvi1:
Just for info the main fix was in pull-request #448.

kasparsuvi1 (approved):

Issue #446, opened 2025-01-24
[TextArea]: Handles inner/outer value incorrectly


kasparsuvi1:
## Problem
Handles inner/outer value incorrectly

## Solution
Should handle state with the same logic as other form components did in community

## Examples and information
* Usage with defaultValue + onChange
* Usage with value + onChange
* Usage with value + onChangeEvent
* usage with defaultValue + onChangeevent

PR #452, opened 2025-01-27
fix(textarea): fix inner value #446

mart-sessman:
Changed textarea inner value logic, now uses same logic TextField component.

### DIFF
--- a/libs/react-components/src/tedi/components/form/textarea/textarea.spec.tsx
+++ b/libs/react-components/src/tedi/components/form/textarea/textarea.spec.tsx
@@ -40,9 +40,9 @@ describe('TextArea component', () => {
     expect(textarea).toHaveClass('tedi-textarea__input');
   });
 
-  it('calls onChange when the textarea value changes', () => {
+  it('calls onChange when using external value and when the textarea value changes', () => {
     const handleChange = jest.fn();
-    render(<TextArea {...defaultProps} onChange={handleChange} />);
+    render(<TextArea {...defaultProps} value="Value" onChange={handleChange} />);
     const textarea = screen.getByPlaceholderText(/enter text/i);
     fireEvent.change(textarea, { target: { value: 'New Value' } });
     expect(handleChange).toHaveBeenCalledTimes(1);
--- a/libs/react-components/src/tedi/components/form/textarea/textarea.spec.tsx
+++ b/libs/react-components/src/tedi/components/form/textarea/textarea.spec.tsx
@@ -21,7 +21,6 @@ describe('TextArea component', () => {
     label: 'Test Label',
     placeholder: 'Enter text...',
     name: 'testTextarea',
-    onChange: jest.fn(),
   };
 
   it('renders the TextArea with default properties', () => {
@@ -50,10 +49,9 @@ describe('TextArea component', () => {
   });
 
   it('displays error when char count is over character limit', () => {
-    const handleChange = jest.fn();
     const charLimit = 10;
     const newText = 'This text is too long';
-    render(<TextArea {...defaultProps} characterLimit={charLimit} onChange={handleChange} />);
+    render(<TextArea {...defaultProps} characterLimit={charLimit} />);
     const textarea = screen.getByPlaceholderText(/enter text/i);
     fireEvent.change(textarea, { target: { value: newText } });
     const error = screen.getByText(`${newText.length}/${charLimit}`);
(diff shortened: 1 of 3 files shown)


airikej (approved):

libs/react-components/src/tedi/components/form/textarea/textarea.tsx
```diff
@@ -13,17 +13,18 @@ export interface TextAreaProps extends TextFieldProps {
...
-
-  const handleInputChange = (inputValue: string) => {
-    setValue(inputValue);
-
-    if (onChange) {
-      onChange(inputValue);
-    }
-  };
+  const { className, helper = [], characterLimit, onChange, value: externalValue, defaultValue, ...rest } = props;
+  const [innerValue, setInnerValue] = React.useState(externalValue ?? defaultValue ?? '');
```
kasparsuvi1, line 17:
```suggestion
  const [innerValue, setInnerValue] = React.useState(defaultValue ?? '');
```
If innerValue is used the externalValue should not be used.

libs/react-components/src/tedi/components/form/textarea/textarea.tsx
```diff
@@ -13,17 +13,18 @@ export interface TextAreaProps extends TextFieldProps {
...
+
+  const handleInputChange = React.useCallback(
+    (inputValue: string) => {
+      setInnerValue(inputValue);
```
kasparsuvi1, line 21: If externalValue and onChange or onChangeEvent is passed the state update should not happen.

kasparsuvi1:
Props still have onChangeEvent, but its not used?

kasparsuvi1:
Add test that check the same bugs that are fixed

kasparsuvi1 (changes requested):

libs/react-components/src/tedi/components/form/textarea/textarea.tsx
mart-sessman: Removed externalValue from initial state.

libs/react-components/src/tedi/components/form/textarea/textarea.tsx
mart-sessman: Added check if externalValue is passed, then just call onChange. (notice onChangeEvent is passed to TextField component with ...rest)
If not, then update inner state.

libs/react-components/src/tedi/components/form/textarea/textarea.tsx
```diff
@@ -13,17 +13,21 @@ export interface TextAreaProps extends TextFieldProps {
...
+  const handleInputChange = React.useCallback(
+    (inputValue: string) => {
+      if (externalValue) {
+        onChange?.(inputValue);
+      } else {
+        setInnerValue(inputValue);
+      }
+    },
+    [externalValue, onChange]
+  );
```
kasparsuvi1, line 28: OnChange should be called always.
```suggestion
  const handleInputChange = React.useCallback(
    (inputValue: string) => {
      if (!externalValue && !(onChange || onChangeEvent) {
        setInnerValue(inputValue);
      }
      onChange?.(inputValue);
    },
    [externalValue, onChange]
  );
```

kasparsuvi1:
Missing tests of the bugs we had:
* defaultValue is not taken into account
* Outer value example with onChange and value properties. TextArea should have correct value after change (Taken from outer state)

mart-sessman:
Added tests.
Added Story for value and onChange event.

libs/react-components/src/tedi/components/form/textarea/textarea.spec.tsx
```diff
@@ -91,4 +83,62 @@ describe('TextArea component', () => {
...
+
+    expect(handleChange).toHaveBeenCalledTimes(1);
+    expect(handleChange).toHaveBeenCalledWith(newValue);
+    expect(textarea).toHaveValue(initialValue);
+
+    render(<TextArea {...defaultProps} value={newValue} onChange={handleChange} placeholder="Enter text" />);
+    const allAreas = screen.getAllByRole('textbox');
+    const newTextarea = allAreas[allAreas.length - 1];
+    expect(newTextarea).toHaveValue(newValue);
+  });
```
kasparsuvi1, line 122: Not sure if it's the best way to test controlled inputs. Rerendering component doesn't directly test, that the value will change correctly. Maybe in future we should consider something like this: https://github.com/testing-library/user-event/issues/549#issuecomment-768271555

kasparsuvi1 (approved):

Issue #422, opened 2025-01-19
[Community/Layout/Sidenav]: Multilevel sidenav


iljakumlander:
## Problem
In development of e-ehitus/ehr new design there is a need to have expandable/collapsable menu items like in this mockup:
https://www.figma.com/design/wTwDyiSJKgGb1leRF39mCU/e-ehitus-Style-2.0?node-id=28768-27389&m=dev

## Solution
With change of `SideNavItem` type and adding `Collapse` element to `sidenav-item`, and adjusting styles.
```tsx
export type SideNavItem<C extends React.ElementType = 'a'> = AnchorProps<C> & {
  /**
   * Icon of the item
   */
  icon?: string | IconProps;
  /**
   * Submenu items
   */
  subItems?: SideNavItem<C>[];
};
```
Some changes needed to `Collapse` element both style and markup-wise.

## Examples and information
- Example implementation: https://storybook.ehr.ee/?path=/story/community-layout-sidenav--colapsable
- Design: https://www.figma.com/design/wTwDyiSJKgGb1leRF39mCU/e-ehitus-Style-2.0?node-id=28768-27389&m=dev

PR #459, opened 2025-02-01
fix(collapse): prevent icon rotation in nested collapses #422

iljakumlander:
- Fix icon rotation inheritance in nested collapse components
- Update selector specificity to only target direct child collapse titles
- Add `data-name="collapse-content" for better` identification
- Add `className="collapse__col--icon"` for icon column styling
- Resolves unintended icon rotations in multilevel menu structures

https://iljakumlander.github.io/tedi-design-system/fix/424-collapse-inside-collapse/react/?path=/docs/community-collapse--docs

### DIFF
--- a/libs/react-components/src/community/components/collapse/collapse.tsx
+++ b/libs/react-components/src/community/components/collapse/collapse.tsx
@@ -96,11 +96,7 @@ export const Collapse = (props: CollapseProps): JSX.Element => {
     }
   };
 
-  const renderContent = (
-    <div className={styles['collapse__content']} data-name="collapse-content">
-      {children}
-    </div>
-  );
+  const renderContent = <div className={styles['collapse__content']}>{children}</div>;
 
   return (
     <div data-name="collapse" {...rest} className={BEM}>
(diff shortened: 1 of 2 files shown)


libs/react-components/src/community/components/collapse/collapse.tsx
```diff
@@ -111,7 +115,7 @@ export const Collapse = (props: CollapseProps): JSX.Element => {
       >
         <Row justifyContent="between" alignItems="center" wrap="nowrap" {...titleRowProps} element="span">
           {title && <Col>{title}</Col>}
-          <Col width="auto">
+          <Col width="auto" className="collapse__col--icon">
```
HarlesPilter, line 118: It's not preferred to encourage external styling which adding this fixed classname does, because it is very fragile. If we decide internally to change this classname or change some css styles, then styles applied by the app might break. Can you give an example what kind of issue this will solve?

iljakumlander: It is required for `Multilevel sidenav` #422 (PR #460) as it is hard otherwise address icon colors in `Collapse` component

HarlesPilter: then there should be prop to allow passing IconProps to the underling icon. Similar to `titleRowProps`

HarlesPilter (approved):

Issue #441, opened 2025-01-22
[Community/Layout/Header]: Custom content support for public header type


iljakumlander:
## Problem
Currently, custom content rendering is limited to system header type only, which restricts applications from displaying `HeaderContent` components in public headers. This limitation prevents consistent content display capabilities across different header types and applications.

## Solution
Implement support for custom content in public headers through the following changes:
- Introduce new `enablePublicCustomContent` prop (defaulting to `false`) to control custom content visibility in public headers
- Replace and deprecate `showSystemCustomContent` prop with more generic `showCustomContent`

This ensures unified content visibility control across both system and public headers using breakpoint-based display logic

## Examples and information
```diff
-            {headerType === 'system' &&
-              renderSystemCustomContent &&
+            {(headerType === 'system' || (enablePublicCustomContent && !shouldBreakToHeader)) &&
+              renderCustomContent &&
```

PR #461, opened 2025-02-01
feat(header): add public header custom content support #441

iljakumlander:
- Add new `enablePublicCustomContent` prop (default: `false`)
- Deprecate `showSystemCustomContent` in favor of `showCustomContent`
- Update documentation for public header content support and prop changes

https://iljakumlander.github.io/tedi-design-system/feat/441-custom-content-support-for-public-header/react/?path=/docs/community-layout--docs

### DIFF
--- a/libs/react-components/src/community/components/layout/header/header.stories.tsx
+++ b/libs/react-components/src/community/components/layout/header/header.stories.tsx
@@ -77,6 +77,34 @@ export const Public: Story = {
   },
 };
 
+/**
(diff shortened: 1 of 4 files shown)


kasparsuvi1:
I think there should be atleast one example story, where public header is used with customContent. And maybe also document that it's not meant to be used with sideNav items? Because i assume it would break the layout.
Otherwise the changes seem 👍

HarlesPilter (approved):
LGM

kasparsuvi1 (approved):

airikej (approved):

Issue #210, opened 2024-10-02
Tooltip


airikej:
https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=5797-117363&t=pQAO0g4IeKdgAGJz-4

[Figma](https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/TEDI-READY-(work-in-progress)?node-id=3125-29823&m=dev)

Current dependency:
- Card

Idea is to break Tooltip and Popover into 2 components:
- Tooltips are typically lightweight, designed for small contextual hints.
- Popovers tend to be more complex, involving richer content like forms, buttons, or additional navigation.

TooltipProps:

| Name      | Required/Optional | Description                                             | Type                      | Possible Values          | Default  |
|-----------|-------------------|---------------------------------------------------------|---------------------------|--------------------------|----------|
| `content` | Required          | The content to display inside the tooltip.             | `React.ReactNode`         |                          |          |
| `placement` | Optional         | Position of the tooltip relative to its trigger.       | `'top' \| 'bottom' \| 'left' \| 'right'` | `top`, `bottom`, `left`, `right` | `'top'`  |
| `trigger` | Optional          | Event that triggers the tooltip.                       | `'hover' \| 'focus' \| 'click'` | `hover`, `focus`, `click` | `'hover'` |
| `children` | Required         | The element that triggers the tooltip when interacted. | `React.ReactNode`         |                          |          |

Add:
- unit test
- breakpoint props
- documentation to ZeroHeight

tarmokullas:
Edastan tiimiliikme poolt tulnud kommentaari. Video on küll hägune kuna githubi jaoks tuli fail väiksemaks teha. Peamine point on siiski videost arussaadav.

===

https://github.com/user-attachments/assets/3d913c00-3c8b-450f-9f2c-5807ce39fdc3

> Pööra siis tähelepanu sellele, mis toimub DOMis kui tooltip nähtavale tuleb - kõikidele DOMi elementidele kirjutatakse mingi atribuut juurde. Ning kui tooltip peitu läheb, võetakse need uuesti ära. See ei kõla üldse optimaalse lahendusena.

[...]

PR #471, opened 2025-02-05
feat(tooltip): TEDI-Ready component #210

mart-sessman:


libs/react-components/.storybook/storybook-decorator.tsx
```diff
@@ -16,7 +17,9 @@ interface StorybookDecoratorProps {
 
 const StorybookDecorator = ({ children, locale = 'en', ...rest }: StorybookDecoratorProps) => (
   <LabelProvider locale={locale} {...rest}>
-    <AccessibilityProvider>{children}</AccessibilityProvider>
+    <AccessibilityProvider>
```
kasparsuvi1, line 20: Getting-started documentation should be updated, that all applications need to wrap inside `AccessibilityProvider` if using Tedi tooltip. Also same info would be good to see under tooltip page.

libs/react-components/src/tedi/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,243 @@
...
+        context,
+        placement,
+      }}
+    >
+      {children}
+    </TooltipContext.Provider>
+  );
+}
+
+Tooltip.Provider = TooltipProvider;
```
kasparsuvi1, line 240: Is it gonna be like this for every new component? Currently we don't use this syntax. Imo it's not needed.

mart-sessman: It's improvement for developers. With this developer needs to import only one component, and get autocomplete from TypeScript for all child components.
Sure it's not needed, but also does not cause any harm.
This needs more discussion between developers I think.

libs/react-components/src/tedi/components/tooltip/tooltip-content.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { children, maxWidth = 'medium', className } = props;
+  const { open, x, y, strategy, focusManager, floating, arrowRef, getFloatingProps, placement, context } =
+    useContext(TooltipContext);
+
+  if (!open) return null;
+
+  return (
+    <FloatingPortal data-name="tooltip">
+      <FloatingFocusManager {...focusManager} context={context}>
+        <div
```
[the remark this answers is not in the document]
mart-sessman: No, we should not use Card inside Tooltip.
For this behaviour Popover component should be used instead (not developed yet).

Tooltip PR will stay open until Popover component is also ready.

libs/react-components/src/tedi/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,243 @@
...
+   * @default tooltip
+   */
+  role?: UseRoleProps['role'];
+  /**
+   * Offset of tooltip.<br />
+   * Used to align HeaderDropdown with bottom of the Header.<br />
+   * By default uses Tooltip.Provider values.
+   * @default GAP + ARROW_HEIGHT (3px + 7px)
+   */
+  offset?: OffsetOptions;
```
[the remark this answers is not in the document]
mart-sessman: Tooltip will be replaced with Popover in Header component.
But I don't understand why offset needs to be removed.
There can be situations in future, where custom offset is needed.

airikej:
@mart-sessman can you add the Tooltip usage documentation to Zeroheight?

airikej (approved):

Issue #260, opened 2024-10-28
Popover


airikej:
https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=5797-117364&t=pQAO0g4IeKdgAGJz-4

PR #491, opened 2025-02-17
feat(popover): TEDI-Ready component development #260

mart-sessman:


libs/react-components/src/tedi/components/tooltip/tooltip-content.tsx
```diff
@@ -1,59 +1,31 @@
...
   /**
-   * Tooltip max width.
+   * Popover width.
    * @default medium
    */
   maxWidth?: 'none' | 'small' | 'medium' | 'large';
 }
 
-export const TooltipContent = (props: TooltipContentProps) => {
+export function TooltipContent(props: TooltipContentProps) {
```
airikej, line 18: What is the reason of turning const into function?

airikej (approved):

PR #506, opened 2025-02-25
feat(colors): TEDI-Ready Angular development #499

mart-sessman:


libs/angular-components/.storybook/preview.js
```diff
@@ -1,10 +1,49 @@
...
+      values: [
+        { name: "default", value: "var(--color-bg-default)" },
+        { name: "muted", value: "var(--color-bg-muted)" },
+        { name: "subtle", value: "var(--color-bg-subtle)" },
+        { name: "disabled", value: "var(--color-bg-disabled)" },
+        { name: "black", value: "var(--color-black)" },
+        { name: "inverted", value: "var(--color-bg-inverted)" },
+        {
+          name: "inverted-contrast",
+          value: "var(--color-bg-inverted-contrast)",
```
airikej, line 16: Can we use TEDI variables here instead - we should of course fix that in React storybook config too, but since we can add more backgrounds when need comes, we can remove those that use old community variables
currenly i would leave only:
`{ name: "brand", value: "var(--primary-600)" },`

libs/angular-components/src/docs/colors/colors.component.ts
```diff
@@ -0,0 +1,79 @@
...
+    `
+      .color-card {
+        min-width: 50px;
+        min-height: 50px;
+        border-radius: 4px;
+        margin-bottom: 8px;
+      }
+      .color__example__wrapper {
+        margin-bottom: 20px;
+      }
```
airikej, line 53: Is it possible to import 

`import '../../tedi-core/src/tedi-storybook-styles.scss';` 

to preview.js and use the styles from there? then you don't have you write inline styles here

airikej (approved):

Issue #236, opened 2024-10-21
Checkbox


airikej:
https://github.com/TEHIK-EE/tedi-design-system/issues/122

[Figma](https://www.figma.com/design/jWiRIXhHRxwVdMSimKX2FF/%E2%9D%97TEDI-READY%E2%9D%97(follow-the-badges-in-header-to-see-which-components-are-ready-(designed%2Fdeveloped))?node-id=4228-72934&m=dev)

Dependencies:
- Grid
- Icon
- Tooltip

Changes:
- No API changes
- Add unit test
- Check WCAG

zalizko:
Hello. I'll copy here my bug report from Slack channel (https://tehikustiilijuhis.slack.com/archives/C045VLXAXD3/p1733953510715499):

I found that currently <Check> component from comminity version of TEDI have a bug.
When form was hidden and enabled again - form reset don't work as expected.

Sandbox: https://codesandbox.io/p/sandbox/react-fiddle-forked-9ztst4.
Steps to reproduce:
1. click checkbox
2. click ‘Toggle’ twice (close and open form again)
3. click ‘Reset’

Nothing happens.

Also the following error occurred in the console: `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

PR #515, opened 2025-02-26
feat(checkbox): TEDI-Ready Checkbox component development #236

airikej:


libs/react-components/src/tedi/components/form/checkbox/checkbox.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (input) {
+      fireEvent.click(input);
+    }
+
+    rerender(<Checkbox id="check-id" label="Check Label" value="check-value" name="check-group" defaultChecked />);
+
+    expect(input).toBeChecked();
+  });
+
+  it('does not change state when clicked if controlled', () => {
```
mart-sessman, line 110: Why input does not change state when controlled?
It should still change state, but it is just controlled input.

mart-sessman (changes requested):

libs/react-components/src/tedi/components/form/checkbox/checkbox.spec.tsx
airikej: A controlled component does not manage its own state - it depends on the parent to update it. If the parent doesn’t update checked, the checkbox stays in its initial state, even after a user interaction.

mart-sessman (approved):

PR #517, opened 2025-02-26
feat(button): Add TEDI Angular button #516

pesarv:
https://github.com/TEHIK-EE/tedi-design-system/issues/516

### DIFF
--- a/libs/angular-components/package.json
+++ b/libs/angular-components/package.json
@@ -36,8 +36,6 @@
     "@angular/platform-browser-dynamic": "~18.2.0",
     "@angular/router": "~18.2.0",
     "@tehik-ee/tedi-core": "^1.5.0",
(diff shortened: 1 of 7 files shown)


pesarv:
[https://pesarv.github.io/tedi-design-system/feat/516-button/angular/?path=/docs/tedi-ready-angular-buttons-button--docs](https://pesarv.github.io/tedi-design-system/feat/516-button/angular/?path=/docs/tedi-ready-angular-buttons-button--docs)

mart-sessman:
Can you move this component and all the related files to community folder?
It currently does not exist, but you can create it in src folder.
And also can you add parent folder to button?

So your button files would be in: "src/community/buttons/button" folder.
In future, if other button components are developed, then we can move these into this "buttons" folder.

PR #547, opened 2025-03-12
feat(typography): TEDI-Ready Angular Typography components #507

mart-sessman:


libs/angular-components/src/tedi/components/typography/text/text.component.ts
```diff
@@ -0,0 +1,121 @@
...
+export class TextComponent implements TextProps {
+  class = input<string | undefined>(undefined);
+  id = input<string | undefined>(undefined);
+  tabIndex = input<number | undefined>(undefined);
+  element = input<TextElement | undefined>("p");
+  modifiers = input<TextModifiers[] | TextModifiers | undefined>(undefined);
+  color = input<TextColor | undefined>("primary");
+
+  classes = computed(() => {
+    const modifiersValue = this.modifiers();
```
airikej, line 102: I think this should default to an empty array if not set 

`this.modifiers() ?? []`

airikej: Same for Heading component

mart-sessman: There is fallback to empty array few lines after it in both components.

airikej (approved):

Issue #544, opened 2025-03-11
Icon


mart-sessman:

mart-sessman:
| Name | Description | Type | Default |
|----------|----------|----------|----------|
name | Name of material icon https://fonts.google.com/icons | string | - |
class | Additional class to style the icon | string | - |
size | Icon size | 8 \| 12 \| 16 \| 18 \| 24 \| 36 \| 48 | 24 |
color | Icon color | "primary" \| "secondary" \| "tertiary" \| "warning" \| "white" \| "success" \| "danger" \| "brand" \| "brand-dark" \| "warning-dark" | "primary" |
background | Icon background color | "primary" \| "secondary" \| "brand-primary" \| "brand-secondary" | - |
filled | Is icon filled? | boolean | false |
label | Icons label for screen-readers. If omitted then the icon is hidden for screen-readers. | string | - |

PR #561, opened 2025-03-14
feat(icon): TEDI-Ready Angular icon component #544

mart-sessman:


libs/angular-components/src/tedi/components/base/icon/icon.stories.ts
```diff
@@ -0,0 +1,393 @@
...
+};
+
+export const UsedInsideText: StoryObj<IconComponent> = {
+  render: (args) => ({
+    props: {
+      ...args,
+    },
+    template: `
+        <div class="example-list padding-14-16">
+            <tedi-heading class="display-flex gap-1">
```
airikej, line 342: The behaviour of Icon inside Text component was that Icon inherits Text components size always and is displayed inline if you see the React example

airikej (approved):

Issue #579, opened 2025-03-18
Card: create community card


pesarv:
### Card
| Prop | Description | Value |
| --- | --- | --- |
| borderless | Removes border from card | boolean |
| spacing | Adds padding to header and content | "xs" \| "sm" \| "md" \| "lg" \| "none" |
| accentBorder | Adds thicker left border | "info" \| "success" \| "warning" \| "danger" |
| selected | Adds selected border to card | boolean |
| background | Changes background for card | card color |

### Card row
By default card contents flow in column, rows can be added for side by side contents
| Prop | Description | Value |
| --- | --- | --- |
| hasSeparator | Adds border under every row | boolean |

### Card content
Card content blocks to be used inside card or card rows
| Prop | Description | Value |
| --- | --- | --- |
| hasSeparator | Inside card adds border under the content, inside card row adds border right of the content | boolean |
| autoWidth | Can be used inside card rows. Changes flex grow to 0 to take up minimal space | boolean |
| timeline | Creates timeline right of the content block. Maybe should be removed/swapped for Separator component once it has been developed like in React  | boolean |
| background | Changes background for card content block | card color |

### Card header
| Prop | Description | Value |
| --- | --- | --- |
| variant | Card header variant | "primary" \| "secondary" \| "tertiary" \| "brand" \| "brand-dark" |

PR #604, opened 2025-03-25
feat(card): communtity card #579

pesarv:
https://github.com/TEHIK-EE/tedi-design-system/issues/579
https://pesarv.github.io/tedi-design-system/feat/579-card/angular/?path=/docs/community-angular-card--docs


pesarv:
Storybooki docsi veel lisan aga kompod võiks kõik olemas olla

mart-sessman:
We changed folder structure, related to fixing some stuff when packaging angular components.
Community folder was before: /src/community
Now it is /community

Can you move your files to there?
Also index.ts where exports were declared is now public-api.ts

mart-sessman:
Why are there underscore's in front of css variables?
Like this: "--_card-padding-vertical"

Have not seen this kind of naming convention before, just asking, is there a reason?
I would remove these, since we dont name our css variables this way, lets try to keep similar patterns as before.

mart-sessman:
Can you create Issue in our repository and then create branch there?
And then lets merge this PR to that new branch. So we can also run, review and test it locally without cloning your fork repository every time and switching to your feature branch.
And for future, could you also create issues in our repository and create branches under the issue, after which you can create PR to "rc" branch.
This saves time for us reviewing it.

pesarv:
> Why are there underscore's in front of css variables? Like this: "--_card-padding-vertical"
> 
> Have not seen this kind of naming convention before, just asking, is there a reason? I would remove these, since we dont name our css variables this way, lets try to keep similar patterns as before.

It's a naming convention for "private"/component scoped css variables. At our first meeting with TEDI team we agreed to using this pattern. It makes it easier to override variables in the application (less specific selectors, easier names etc.)

PR #650, opened 2025-04-02
Feature/637 form label

erkidorbek:
**issue:** [Add community form label](https://github.com/TEHIK-EE/tedi-design-system/issues/637)

**github pages:** https://erkidorbek.github.io/tedi-design-system/feature/637-form-label/angular/?path=/docs/community-angular-form-label--docs

### DIFF
--- a/libs/angular-components/community/components/form/index.ts
+++ b/libs/angular-components/community/components/form/index.ts
@@ -1,0 +1,1 @@
+export * from "./label/label.component";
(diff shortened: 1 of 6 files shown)


mart-sessman:
It would be create if you can create issue in our repository and then create branch from there.
Currently we have to fork your repository, switch to some branch and then run it, every time when reviewing.

mart-sessman (changes requested):

PR #653, opened 2025-04-03
feat(link): TEDI-Angular Community Link #621

mart-sessman:


libs/angular-components/community/components/navigation/link/link.component.html
```diff
@@ -0,0 +1,71 @@
+<a
+  *ngIf="breakpointInputs().href; else routerLink"
+  [class]="classes()"
+  [attr.href]="breakpointInputs().href"
```
priitpu, line 4: Different destinations based on screen size?

libs/angular-components/community/components/navigation/link/link.component.html
```diff
@@ -0,0 +1,71 @@
...
+  [attr.target]="breakpointInputs().target"
+  [attr.rel]="breakpointInputs().rel"
+  [attr.id]="breakpointInputs().id"
+  [attr.tabIndex]="0"
+>
+  <tedi-icon
```
priitpu, line 10: Should a link component be coupled with the icon component? Can't this be handled by the consumer side by using it `<a ...inputs><tedi-icon ...inputs />content</a>`. Gaps and such could be handled by link targeting an icon.

libs/angular-components/community/components/navigation/link/link.component.ts
```diff
@@ -0,0 +1,200 @@
...
+
+@Component({
+  selector: "tedi-link",
+  standalone: true,
+  templateUrl: "./link.component.html",
+  styleUrl: "./link.component.scss",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgIf, NgTemplateOutlet, IconComponent, RouterLink],
+})
+export class LinkComponent implements BreakpointInputs<LinkInputs> {
```
priitpu, line 122: I feel like creating the component by using an attribute selector would be better so you don't have to expose all of all anchor attributes or pass them down to the `a` tag

erkidorbek:
cant we just use default link (<a> with tedi-link directive? We just basically need to add and style [size] and [underline], or am i missing something crucial here?

Icon can be managed inside <a> tag, it doesn't need to be baked into the component.

angular material atleast does it like that.

```
<div class="example-button-container">
    <a mat-fab extended routerLink=".">
        <mat-icon>favorite</mat-icon>
        Link
    </a>
</div>
```

So our equivalent should be something like

```
<a tedi-link ...(routerlink etc stuff)>
    <tedi-icon blablabla />
    My cool link string
</a>
```

(https://material.angular.io/components/button/examples last example for mat-fab, but doable for tedi link styling also)

mart-sessman:
We today just talked about using attribute selector, this will get changed.

libs/angular-components/community/components/navigation/link/link.component.html
mart-sessman: This will be removed, switching to attribute selector, these inputs are no longer needed.

libs/angular-components/community/components/navigation/link/link.component.html
mart-sessman: Fair question.
When consumer adds tedi-icon component itself, then there is also some styling needed.
By default link has underline, and now this tedi-icon will also have underline, but this is against the design.
Also tedi-icon should inherit it's font-size from the link, which means overriding default icon size.
That's why iconLeft and iconRight inputs were added to tedi-link component.

herrsiim:
> .... Regarding to the icon component inside link component, yes this can be removed, but this means consumer application developers need to do more work. Some additional styling is needed (remove underline from icon, add gap, add correct font-size). Right now I did not remove iconLeft and iconRight inputs from component, but the discussion is still open.

I’d lean towards keeping iconLeft and iconRight within the component, especially if the styling is handled internally via the component’s CSS. It simplifies things for consumers by making it easier to add icons to links without needing extra setup or custom styling on their side.

libs/angular-components/community/components/navigation/link/link.component.ts
herrsiim: Attribute selector solution looks great!

PR #661, opened 2025-04-03
feat(feedback-text): Community Angular feedback-text component #654

mart-sessman:


libs/angular-components/community/components/form/feedback-text/feedback-text.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  computed,
+  input,
+  ViewEncapsulation,
+} from "@angular/core";
+
+export type FeedbackTextType = "hint" | "valid" | "error";
+export type FeedbackTextPosition = "left" | "right";
+
+@Component({
+  selector: "div[tedi-feedback-text]",
```
herrsiim, line 13: Would this also work? I might be wrong, but why wouldn't we want to use <edi-feedback-text></edi-feedback-text> approach here?
```javascript
selector: "tedi-feedback-text, div[tedi-feedback-text]"
```

mart-sessman: Since output will be single HTML element, I don't see a reason why use custom selector, which wraps it with host element.

herrsiim (approved):

Issue #552, opened 2025-03-12
[Accordion]: Create community accordion component


pesarv:
### Accordion
Wrapper of accordions which can control their opened state
| Prop | Description | Value |
| --- | --- | --- |
| defaultOpenItems | id-s of items that should be opened by default | string[] |
| singleOpen | Whether only one accordion item can be opened at once | boolean |

### Accordion item
Single collapsible card
| Prop | Description | Value |
| --- | --- | --- |
| id | Accordion item id | string |
| selected | Whether accordion item is selected| boolean |

### Accordion icon
Component to be used inside accordion items with icons. Creates left side content with secondary background as default and add columns to accordion item.
| Prop | Description | Value |
| --- | --- | --- |
| background | Changes background for accordion icon content block | card color |

### Accordion item header
Header of accordion item
| Prop | Description | Value |
| --- | --- | --- |
| indicator | Whether accordion header should have indicator arrow | boolean |
| openText | Open button text. Visible if both openText and closeText are provided | string |
| closeText | Close button text. Visible if both openText and closeText are provided | string |
| variant | Accordion header variant. Same as card header variant  | card header variant |
| background | Changes background for accordion header | card color |
| indicatorColor | Changes indicator arrow color | icon color |

### Accordion item content
Content of accordion item
| Prop | Description | Value |
| --- | --- | --- |
| background | Changes background for accordion content | card color |

tobre6:
There should be option to have different header color for active state:

![Image](https://github.com/user-attachments/assets/c23e8007-7f11-4536-a8ff-b87ba23a7b00)

PR #662, opened 2025-04-03
feat(accordion): create community accordion component #552

pesarv:
Community accordion


pesarv:
In Figma accordion is under buttons. Does in make sense or should I rather move it into a separate folder?

tobre6:
> In Figma accordion is under buttons. Does in make sense or should I rather move it into a separate folder?

@karolin-kivisikk. I think the question is more than resonable. Maybe you can elaborate

karolin-kivisikk:
> In Figma accordion is under buttons. Does it make sense or should I rather move it into a separate folder?

Good question I don't remember why it was left under buttons but I moved it under Cards folder/section in Figma and Zeroheight.

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  &-indicator {
+    transition: transform 0.3s;
+  }
+
+  & &-toggle-btn {
+    --_btn-padding-y: 0;
+  }
+
+  &--opened > &-indicator,
```
priitpu, line 25: What flavour of Block-Element-Modifier is this? 🤔

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.ts
```diff
@@ -0,0 +1,84 @@
...
+      return this.opened() ? this.closeText() : this.openText();
+    }
+    return;
+  });
+
+  indicatorIconColor = computed(() => {
+    if (this.indicatorColor()) {
+      return this.indicatorColor()!;
+    }
+    return ["brand", "brand-dark"].includes(this.variant() as CardHeaderVariant)
```
priitpu, line 76: Shouldn't this be set as a constant? 🤔

libs/angular-components/community/components/buttons/accordion/accordion/accordion.component.ts
```diff
@@ -0,0 +1,67 @@
...
+              otherItem.close();
+            }
+          });
+        }
+      });
+    });
+  }
+
+  ngOnInit() {
+    this.openDefaultOpenItems();
```
priitpu, line 62: What happens if i set defaultOpenItems to an array of id's and singleOpen as true?

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.html
```diff
@@ -0,0 +1,38 @@
...
+  @if (indicator()) {
+    <tedi-icon
+      name="keyboard_arrow_down"
+      class="tedi-accordion-item-header-indicator"
+      [size]="18"
+      [color]="indicatorIconColor()"
+    />
+  }
+  @if (toggleButtonText()) {
+    <button
```
mart-sessman, line 22: Button inside button element is not valid HTML.
I would say that change first button to div, add tabIndex="0" and role="button" and this should solve it.

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  gap: var(--_accordion-header-spacing);
+  align-items: center;
+  justify-content: space-between;
+
+  [tedi-accordion-header-end] {
+    margin-left: auto;
+  }
+
+  &-indicator {
+    transition: transform 0.3s;
```
mart-sessman, line 18: Perhaps &-indicator --> &__indicator ?
Same for toggle-btn

libs/angular-components/community/components/buttons/accordion/accordion-item/accordion-item.component.html
```diff
@@ -0,0 +1,6 @@
+<tedi-card
+  [selected]="selected()"
+  class="tedi-accordion-item__card"
+  [class.tedi-accordion-item__card--expanded]="opened()"
+  ><ng-content></ng-content
```
mart-sessman, line 5: Can you fix line endings and new lines in this file and in other files, where you have html? (.html files and in stories)

So it would be:
```
<tedi-card
   [selected]="selected()"
   class="tedi-accordion-item__card"
   [class.tedi-accordion-item__card--expanded]="opened()"
>
   <ng-content></ng-content>
</tedi-card>
```


I will add some sort of linter which will fix this automatically in future.

libs/angular-components/community/components/buttons/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,227 @@
...
+          <tedi-text color="secondary">
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+          </tedi-text>
+        </tedi-accordion-item-content>
+      </tedi-accordion-item>
+
+      <tedi-accordion-item id="accordion-3" [selected]="true">
+        <tedi-accordion-item-header [indicator]="true">
+          <tedi-text color="secondary">Title</tedi-text>
+          <button
```
mart-sessman, line 131: In design there is possible to select between different items, can we add this here aswell?
Currently this is always selected.

Example in react:
https://github.tehik.ee/tedi-design-system/rc/react/?path=/docs/community-accordion--docs#selected

mart-sessman (changes requested):

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.html
priitpu: Seeing the use case for this, (select "button in accordion header") this shouldnt be a button at all, should probably be a radio group or checkboxes. Buttons do not have state (unless you give them state using aria-checkbox or some other HTML compatible way).

priitpu: A native radio group would also handle this in a browser friendly and accessible way

priitpu: Should consider considering this feature out of scope for the accordion

libs/angular-components/community/components/buttons/accordion/accordion.stories.ts
priitpu: Seeing the use case for this, (select "button" in accordion header) this shouldnt be a button at all, should probably be a styled radio group or checkboxes. Buttons do not have state (unless you give them state using aria-checkbox or some other HTML compatible way). Better terminology would be "pressable" i guess.

A native radio group would also handle this in a browser friendly and accessible way

Should consider considering this feature out of scope for the accordion?

mart-sessman: Yes, I agree with you, that it should be radio group or checkbox, just styled as a button.
Luckily this feature is not inside accordion component itself, it is just ng-content inside item-header story.
I guess this could be removed from the story for now then?
And we can implement this later, when radio group/checkbox components are developed?

libs/angular-components/community/components/buttons/accordion/accordion/accordion.component.ts
pesarv: only first one is opened

libs/angular-components/community/components/buttons/accordion/accordion-item-header/accordion-item-header.component.html
pesarv: By HTML standard there shouldn't be nested interactive elements, so i changed the nested button to span and also removed the selected buttons from stories

libs/angular-components/community/components/buttons/accordion/accordion.stories.ts
pesarv: removed

mart-sessman:
@pesarv I merged RC into this branch and did some fixes/quality changes.
You can look these over, and if everything seems ok, put thumbs up and I will merge it.

pesarv:
> @pesarv I merged RC into this branch and did some fixes/quality changes. You can look these over, and if everything seems ok, put thumbs up and I will merge it.

lgtm

PR #690, opened 2025-04-14
feat(number-field): TEDI Angular Community component #665

mart-sessman:


libs/angular-components/community/components/form/number-field/number-field.component.ts
```diff
@@ -0,0 +1,184 @@
...
+})
+export class NumberFieldComponent implements ControlValueAccessor {
+  /**
+   * The unique identifier for the input element that this label is associated with. This ID should match the input element's id attribute to ensure accessibility.
+   */
+  id = input.required<string>();
+  /**
+   * The text content of the label that describes the input field.
+   */
+  label = input.required<string>();
```
pesarv, line 55: Would be better if label was not required. In our application the label placement will probably be handled by the form field component

libs/angular-components/community/components/form/number-field/number-field.component.scss
```diff
@@ -0,0 +1,180 @@
+@use "@tehik-ee/tedi-core/bootstrap-utility/breakpoints";
+@use "@tehik-ee/tedi-core/mixins";
+
+$input-height: 2.5rem;
```
pesarv, line 4: Would rather use css variables instead of sass variables. Makes it easier to overwrite in the end application

libs/angular-components/community/components/form/number-field/number-field.component.html
```diff
@@ -0,0 +1,75 @@
...
+>
+  <button
+    tedi-button
+    type="button"
+    variant="secondary"
+    class="tedi-number-field__button tedi-number-field__button--decrement"
+    [disabled]="decrementDisabled()"
+    (click)="handleButtonClick('decrement')"
+  >
+    <tedi-icon name="remove" [size]="18" [label]="'Decrement by ' + step()" />
```
pesarv, line 25: Should probably have some way to provide translated labels

libs/angular-components/community/components/form/number-field/number-field.component.scss
mart-sessman: Using css variables.

libs/angular-components/community/components/form/number-field/number-field.component.ts
mart-sessman: Not required anymore.

herrsiim (approved):

pesarv (approved):

PR #710, opened 2025-04-17
682 community radio

pesarv:


herrsiim:
Please review the CSS for the radio dot placment.
image
image

pesarv:
> Please review the CSS for the radio dot placment, it's not centered correctly. image image

Updated it to be centered by flex. On some zoom levels browsers can render it half a pixel off since the dot size is percentage of the radio button itself, but should be better now

libs/angular-components/community/components/form/radio/radio/radio.component.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  disabled = input<boolean>();
+  /*
+   * Whether radio has error.
+   */
+  hasError = input<boolean>();
+  /**
+   * FeedbackText component inputs.
+   */
+  feedbackText = input<any>(); // TODO: change after number input merged
```
mart-sessman, line 55: This can be changed now, and in other component too.

libs/angular-components/community/components/form/radio/radio/radio.component.scss
```diff
@@ -0,0 +1,128 @@
...
+    align-items: center;
+    justify-content: center;
+
+    &::after {
+      content: "";
+      border-radius: 50%;
+      display: none;
+      background: var(--_radio-dot-bg);
+      height: 70%;
+      width: 70%;
```
mart-sessman, line 63: Radio dot still not centered on my screen, maybe using percentages as width and height is not the best solution here.
When using percentages here some screens may render 1px more or less than others, so fixed value would be ideal here.

libs/angular-components/community/components/form/radio/radio/radio.component.scss
```diff
@@ -0,0 +1,128 @@
+@use "@tehik-ee/tedi-core/mixins";
+
+@mixin input-state-styles {
+}
+
+.tedi-radio {
```
mart-sessman, line 6: Currently radio can not be focused with keyboard navigation or missing styles for it.
You can look radio component in react how it works there.

mart-sessman (changes requested):

libs/angular-components/community/components/form/radio/radio.stories.ts
```diff
@@ -0,0 +1,345 @@
...
+    return {
+      props: { ...args, control },
+      template: `
+      <tedi-radio-group
+        id="radio-group"
+        name="radio-story"
+        [formControl]="control"
+        ${argsToTemplate(args)}
+      >
+        <tedi-radio name="test" inputId="radio-1" value="radio-1"
```
mart-sessman, line 101: Currently every story uses same id-s, so when pressing on label in some other story it navigates to first story.
So use different Id-s in stories.

mart-sessman (changes requested):
If focus is fixed, I think we can merge it.

libs/angular-components/community/components/form/radio/radio/radio.component.scss
```diff
@@ -0,0 +1,130 @@
...
+    --_radio-bg: var(--form-checkbox-radio-default-background-error);
+    --_radio-border: var(--form-general-feedback-error-border);
+  }
+
+  &--large {
+    @include mixins.responsive-styles(
+      --_radio-size,
+      form-checkbox-radio-size-large
+    );
+    --_radio-dot-size: 16px;
```
mart-sessman, line 124: Here comes the SASS warning. Include should be after radio-dot-size variable.

mart-sessman (changes requested):

libs/angular-components/community/components/form/radio/radio/radio.component.scss
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  &:has(#{&}__input:checked) {
+    .tedi-radio__indicator::after {
+      display: block;
+    }
+    --_radio-border: var(--form-checkbox-radio-default-border-selected);
+  }
+
+  &:has(#{&}__input:hover, #{&}__input:focus-visible) {
```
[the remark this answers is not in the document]
pesarv: This is how native radio works, it focuses the first radio and then can be navigated with arrows

mart-sessman:
Fix radio cards, so they wont "jump" when selecting/unselecting as we discussed in meeting.

mart-sessman:
There is still some horizontal movement, when hovering radio cards.

https://github.com/user-attachments/assets/f8193176-907f-4654-a981-99de2d1f2385

PR #712, opened 2025-04-22
Feature/640 community select w content projection

erkidorbek:
Github pages:
https://erkidorbek.github.io/tedi-design-system/feature/640-community-select-w-content-projection/angular/?path=/docs/community-angular-form-select--docs

Issue:
https://github.com/TEHIK-EE/tedi-design-system/issues/640

biggest concerns:

- using ControlValueAccessor
- using conent projection with `<ng-content>`
- accessibility with cdkMenu (focuses first element and traps focus on keyboard interaction, etc)


libs/angular-components/community/components/form/select/dropdown-item.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  selected = input<boolean>(false);
+
+  /**
+   * Applies the disabled style to the dropdown item.
+   */
+  disabled = input<boolean>(false);
+
+  modifierClasses = computed(() => {
+    const modifiers = [];
+    if (this.selected()) modifiers.push(`tedi-dropdown-item--active`);
```
pesarv, line 33: for boolean checks can jost directly bind to the host `"[class.tedi-dropdown-item--active]" : "selected()"`

libs/angular-components/community/components/form/select/select-option.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  private elementRef = inject(ElementRef);
+  contentText: string = "";
+
+  isSelected = computed(() => {
+    return this.parent._selectedValue() === this.value();
+  });
+
+  ngAfterContentInit() {
+    // Extract text content from the component to use as label
+    this.contentText = this.elementRef.nativeElement.textContent.trim();
```
pesarv, line 55: maybe not necessary to store it as a class property, just extract it when selecting

pesarv: also check maybe is possible to have it just as a computed signal in the parent? from the SelectOptionComponent contentchildren filter out the selected option and show its label, without explicitly setting it

libs/angular-components/community/components/form/select/dropdown-item.component.ts
erkidorbek: changed

mart-sessman:
Can you provide github pages link?

erkidorbek:
Still some "any" types what linter doesn't like. Not sure how to deal with them rn, but overall should be reviewable.

libs/angular-components/community/components/form/select/select.component.html
```diff
@@ -0,0 +1,65 @@
...
+  >
+    @if (selectedLabel(); as label) {
+      {{ label }}
+    } @else {
+      {{ placeholder() }}
+    }
+  </span>
+
+  @if (_selectedValue()) {
+    <button
```
mart-sessman, line 23: Button inside button is not valid HTML.
Maybe change the outer button to div element and add aria attributes.

mart-sessman (changes requested):

libs/angular-components/community/components/form/select/select-option.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, TemplateRef, input, viewChild } from "@angular/core";
+
+@Component({
+  selector: "tedi-select-option",
+  standalone: true,
+  imports: [],
+  template: `<ng-template><ng-content /></ng-template>`,
```
herrsiim, line 7: <ng-content> on mõeldud sisestuse jaoks, mis toimub komponendi mallis otse, mitte dünaamiliselt.
<ng-template> on dünaamilise sisu jaoks (ehk sisu, mida saab runtime’is luua *ngIf, *ngFor, ngTemplateOutlet jne kaudu).
Neid kahte ei tohiks kombineerida nii.

herrsiim:
Add select to: `community/components/form/index.ts`

libs/angular-components/community/components/form/select/select.component.ts
```diff
@@ -0,0 +1,150 @@
...
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => SelectComponent),
+      multi: true,
+    },
+  ],
+  host: {
+    "[class.tedi-select]": "true",
+  },
+})
+export class SelectComponent
```
herrsiim, line 59: Add to community/components/form/.index.ts

libs/angular-components/community/components/overlay/dropdown-item/dropdown-item.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  styleUrl: "./dropdown-item.component.scss",
+  host: {
+    "[class.tedi-dropdown-item]": "true",
+    "[class.tedi-dropdown-item--active]": "this.selected()",
+    "[class.tedi-dropdown-item--disabled]": "this.disabled()",
+  },
+})
+export class DropdownItemComponent {
```
herrsiim, line 20: Should we include this in `Add to community/components/form/index.ts`?

herrsiim (changes requested):

herrsiim:
Why is `<ng-content>` wrapped inside `<ng-template>`? I don't think this is correct usage for combining them both like this.

erkidorbek:
> Why is `<ng-content>` wrapped inside `<ng-template>`? I don't think this is correct usage for combining them both like this.

https://angular.dev/api/common/NgTemplateOutlet

libs/angular-components/community/components/overlay/dropdown-item/dropdown-item.component.ts
erkidorbek: In figma it'd under "overlay", so maybe i should create new index.ts inside components/overlay/ and then import it into components index.ts?

libs/angular-components/community/components/form/select/select.component.ts
erkidorbek: added entry points

libs/angular-components/community/components/form/select/select.component.scss
```diff
@@ -0,0 +1,58 @@
...
+  &__label {
+    text-align: left;
+    flex-grow: 1;
+
+    &--placeholder {
+      color: var(--_placeholder-color);
+    }
+  }
+
+  &__clear {
```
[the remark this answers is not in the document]
erkidorbek: Created the closing button

libs/angular-components/community/components/overlay/dropdown-item/dropdown-item.component.ts
erkidorbek: added under /overlay

libs/angular-components/community/components/form/select/select.component.ts
```diff
@@ -0,0 +1,150 @@
...
+  // Internal state
+  _selectedValue = signal(null);
+  _disabled = signal<boolean>(false);
+  _width = signal<number>(0);
+  _options = contentChildren(SelectOptionComponent);
+
+  private selectRef = inject(ElementRef);
+
+  // ControlValueAccessor methods
+  private onChange = (value: any): void => {};
```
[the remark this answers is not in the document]
erkidorbek: mat angular has "any" in the value type, since I think the select "value" can also be any type of object. 

image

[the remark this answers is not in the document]
erkidorbek: https://github.com/angular/components/blob/main/src/material/select/select.ts

libs/angular-components/community/components/form/select/select.component.html
```diff
@@ -0,0 +1,66 @@
+<button
+  tedi-input
+  type="button"
+  [state]="state()"
+  [size]="size()"
+  [cdkMenuTriggerFor]="selectDropdown"
+  [disabled]="_disabled()"
```
herrsiim, line 7: If the select component is disabled, then there must not be any highlights when you click on disabled component.

herrsiim (approved):

PR #713, opened 2025-04-22
feat(tooltip): Angular community tooltip  #669

herrsiim:
New `TooltipComponent` using Angular CDK's overlay and portal modules. It supports customizable text, positions (`top`, `bottom`, `left`, `right`), triggers (`hover`, `click`), and widths (`none`, `small`, `medium`, `large`). Includes lifecycle hooks for event handling and overlay management.


libs/angular-components/community/components/overlay/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+  render: (args) => ({
+    props: args,
+    template: `<p tedi-text>Tooltip works even inside a text. Hover over 
+      <tedi-tooltip text="Tooltip content" positions="top">
+        <span #tooltipTrigger tedi-text color="brand">this</span>
+      </tedi-tooltip>
+      text to see the tooltip. 
+    </p>`,
+  }),
+};
```
mart-sessman, line 158: I would add one story with longer text and where you have different maxWidth values.

libs/angular-components/community/components/overlay/tooltip/tooltip.component.ts
```diff
@@ -0,0 +1,209 @@
...
+   * The text to be displayed in the tooltip.
+   */
+  text = input.required<string>();
+  /**
+   * The position of the tooltip relative to the trigger element. If tooltip can't
+   * be positioned in the specified direction, the CDK will try to position the tooltip
+   * in the next direction in positions list.
+   * @default 'top, bottom, left, right'
+   */
+  positions = input<string>("top, bottom, left, right");
```
mart-sessman, line 91: I am wondering if this should be one position only? The position where user tries to render the tooltip.
And if it's not possible then the component handles itself to render tooltip in opposite direction?

libs/angular-components/community/components/overlay/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+          summary: "click | hover",
+        },
+      },
+    },
+  },
+  render: (args) => ({
+    props: args,
+    template: `
+      <tedi-tooltip ${argsToTemplate(args)}>
+        <button #tooltipTrigger tedi-info-button></button>
```
mart-sessman, line 96: Is this #tooltipTrigger always needed to be included?
Or can this be handled internally? So it is added by the tooltip component?

mart-sessman (changes requested):

libs/angular-components/community/components/overlay/tooltip/tooltip.stories.ts
herrsiim: We ran several tests by using `elements = contentChildren<ElementRef>('*')` and `@ContentChildren('*', { read: ElementRef }) elements!: QueryList<ElementRef>` but none of them couldn't catch any HTML elements. Currently using the template refrence variable is the way to go.

libs/angular-components/community/components/overlay/tooltip/tooltip.component.ts
herrsiim: Done

libs/angular-components/community/components/overlay/tooltip/tooltip.stories.ts
herrsiim: Done

libs/angular-components/community/components/overlay/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,183 @@
...
+    text: {
+      control: "text",
+      description: "The text to display in the tooltip.",
+    },
+    position: {
+      control: "text",
+      description:
+        "The position of the tooltip relative to the trigger element. ",
+      defaultValue: {
+        summary: '"top, bottom, left, right"',
```
mart-sessman, line 56: Fix summary.

mart-sessman (changes requested):

PR #722, opened 2025-04-23
feat(collapse): TEDI-Ready Collapse component development #696

airikej:


libs/react-components/src/tedi/components/buttons/collapse/collapse.module.scss
```diff
@@ -0,0 +1,40 @@
+@use '@tehik-ee/tedi-core/mixins';
+
+$btn-height: 2rem;
```
mart-sessman, line 3: This should be some css variable?

mart-sessman (changes requested):

mart-sessman (approved):

PR #725, opened 2025-04-24
feat(form-textarea): Add stories for textarea #639

erkidorbek:
Using the already created [tedi-input] directive to apply the input styles and [size] and [state] inputs. Created 2 stories: basic textarea story without form and example with label+feedbacktext+formfield under form-field story

**Github pages**: 
- https://erkidorbek.github.io/tedi-design-system/feature/639-community-textarea/angular/?path=/docs/community-angular-form-textarea--docs
- https://erkidorbek.github.io/tedi-design-system/feature/639-community-textarea/angular/?path=/story/community-angular-form-formfield--with-textarea

**Issue**:
https://github.com/TEHIK-EE/tedi-design-system/issues/639


mart-sessman:
Why not create separate textarea component?
Also currently textarea can be resized both horizontally and vertically.
In React this is possible only vertically. And there is minimum size, from which textarea can not be smaller.

I'm not saying that in Angular we should remove horizontal resizing, but the minimum size should be applied.
Maybe create some input for resizing which has values "horizontal", "vertical", "both", "none" ? Just thinking out loud.

libs/angular-components/community/components/form/input/textarea/input-textarea.component.scss
```diff
@@ -0,0 +1,28 @@
+@use "@tehik-ee/tedi-core/mixins";
+
+.tedi-textarea {
+  @include mixins.responsive-styles(--_min-height, form-textarea-min-height);
```
mart-sessman, line 4: Put @include after padding, since it gives SASS deprecation warnings otherwise.
@include mixins should always be last in each block.

But it can before &--resizeX and &--resizeY, because these are new "blocks".

libs/angular-components/community/components/form/input/textarea/input-textarea.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  input,
+  ViewEncapsulation,
+} from "@angular/core";
+import { InputComponent } from "../input.component";
+
+@Component({
+  selector: "[tedi-textarea]",
+  standalone: true,
+  template: "<ng-content />",
+  styleUrls: ["../input.component.scss", "./input-textarea.component.scss"],
```
mart-sessman, line 13: Name textarea component as textarea.component.
I think input- prefix is not needed at the start. Sure, it gets some styles from the input, but it's still separate component.

libs/angular-components/community/components/form/input/textarea/input-textarea.component.scss
```diff
@@ -0,0 +1,28 @@
...
+  );
+  @include mixins.responsive-styles(
+    --_padding-y,
+    form-field-padding-y-md-default
+  );
+
+  width: 100%;
+  min-height: var(--_min-height);
+  padding: var(--_padding-y) var(--_padding-x);
+
```
mart-sessman, line 17: `resize: none;` missing.
If you set both resizeX and rezizeY to false, this textarea can still be resized since it is default.

libs/angular-components/community/components/form/input/textarea/input-textarea.component.scss
mart-sessman: And also I'm wondering, why not use tedi-core variable with mixins?
Why create separate layer of variables?

You could do like this:
```
@include mixins.responsive-styles(min-height, form-textarea-min-height);
@include mixins.responsive-styles(padding, form-field-padding-y-md-default form-field-padding-x-md-default);
```

mart-sessman (changes requested):

libs/angular-components/community/components/form/input/textarea/input-textarea.component.scss
erkidorbek: changed

libs/angular-components/community/components/form/input/textarea/input-textarea.component.ts
erkidorbek: changed

libs/angular-components/community/components/form/input/textarea/input-textarea.component.scss
erkidorbek: changed

mart-sessman:
Last thing: move textarea component from "components/form/input/textarea" to "components/form/textarea".

mart-sessman (approved):

PR #729, opened 2025-04-25
feat(list): Angular list items color input update

herrsiim:
- Added a `color` input property to the `ListComponent` to allow users to specify the bullet color, with a default value of `"brand"`.
- Adjusted padding for list bullets in mobile view.

### DIFF
--- a/libs/angular-components/tedi/components/content/list/list.component.scss
+++ b/libs/angular-components/tedi/components/content/list/list.component.scss
@@ -27,7 +27,7 @@ ul.tedi-list {
 
   li {
     position: relative;
-    padding-left: var(--list-padding-left-level-1-desktop);
+    @include mixins.responsive-styles(padding-left, list-padding-left-level-1);
 
     &::before {
       position: absolute;
--- a/libs/angular-components/tedi/components/content/list/list.component.scss
+++ b/libs/angular-components/tedi/components/content/list/list.component.scss
@@ -23,14 +23,6 @@ $bullet-colors: (
 ul.tedi-list {
   list-style: none;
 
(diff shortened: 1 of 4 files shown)


libs/angular-components/tedi/components/content/list/list.component.ts
```diff
@@ -31,6 +48,10 @@ export class ListComponent {
       classList.push("tedi-list--unstyled");
     }
 
+    if (this.color()) {
```
mart-sessman, line 51: Unnecessary if check, since color is always defined, because you use initialValue with input.
You can do like this:
```const classList = ["tedi-list", `tedi-list--bullet-color-${this.color()}`]```

File: libs/angular-components/tedi/components/content/list/list.component.scss [hunk 2]
```diff
@@ -9,21 +22,19 @@
...
+
   @include mixins.responsive-styles(padding-left, list-padding-left-level-1);
 
   li {
     position: relative;
-    @include mixins.responsive-styles(
-      padding-left,
-      list-padding-left-level-1
-    );
+    padding-left: var(--list-padding-left-level-1-desktop);
```
mart-sessman, line 30: See if list-padding-left-level-1-mobile can be changed to 1rem in TEDI-Core.
If yes, then use current mixin.

mart-sessman (changes requested):

libs/angular-components/tedi/components/content/list/list.component.scss
herrsiim: Will be updated in: 
- #733

Issue #741, opened 2025-04-29
[tedi-core]: responsive-style mixin update


herrsiim:
We are updating the responsive-styles mixin to dynamically combine a base variable (e.g., list-inner-spacing-x) with an additional variable (e.g., list-icon-size) using calc(). This ensures that the generated CSS applies responsive styles across breakpoints while correctly handling both variables for calculations.

PR #742, opened 2025-04-29
feat(tedi-core): simplify responsive-styles #741

herrsiim:
Updating `_mixins.scss`  by removing an unnecessary function by extracting css calc() outside of responsive-style.
Removing ´apply-calculation´ function, as the calculation should be done outside of mixins.


herrsiim:
This PR will stary draft until we can merge: 
- #749

mart-sessman (approved):

Issue #744, opened 2025-04-30
[Alert]: Icon will be partially hidden if text grows larger


airikej:
## Problem

If it has no title and the content text grows larger then the icon will be hidden

PR #746, opened 2025-04-30
fix(alert): prevent icon to be partially hidden if text grows larger #744

airikej:


libs/react-components/src/tedi/components/notifications/alert/alert.tsx
```diff
@@ -102,7 +102,11 @@ export const Alert = (props: AlertProps): JSX.Element | null => {
     const iconProps: IconWithoutBackgroundProps =
       typeof icon === 'string' ? { ...defaultIconProps, name: icon } : { ...defaultIconProps, ...icon };
 
-    return <Icon {...iconProps} />;
+    return (
+      <div className={styles['tedi-alert__icon']}>
+        <Icon {...iconProps} />
```
mart-sessman, line 107: Can't we just add flex-shrink to icon?
And then this wrapper div is not needed?

mart-sessman (changes requested):

mart-sessman (approved):

PR #753, opened 2025-05-01
681 checkbox community checkbox

pesarv:


libs/angular-components/community/components/form/checkbox/checkbox-card-group/checkbox-card-group.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  override direction = input<"row" | "column">("row");
+  /*
+   * Value of card group. Do not use with controlled card group.
+   */
+  value = model<string[]>([]);
+
+  /*
+   * Emits selected checkbox values.
+   */
+  valueChange = output<string[]>();
```
mart-sessman, line 65: Doesn't value = model generate input/output pair, so this valueChange output is not needed?
https://angular.dev/api/core/model

pesarv: yup, correct, removed

mart-sessman:
In Safari browser Radio Card Groups don't work as expected.

1. Whole card does not trigger selection (only label triggers it). In Chrome this works.
2. Radio indicator does not trigger selection


https://github.com/user-attachments/assets/77e07cbf-624b-439a-9d7c-91ded65d0336

pesarv:
> In Safari browser Checkbox Card Group and Radio Card Groups don't work as expected.
> 
> 1. Whole card does not trigger selection (only label triggers it). In Chrome this works.
> 2. Radio indicator does not trigger selection
> 
> EDIT: I made video of Radio Card Groups, but same behaviour is with Checkbox Card Group in Safari.
> 
>  Screen.Recording.2025-05-07.at.12.04.22.mov

apparently safari still doesnt support inset property, fixed

herrsiim (approved):

Issue #575, opened 2025-03-18
Label support


mart-sessman:
## Problem
Missing Label support for Angular components

## Solution
Add Label support

PR #782, opened 2025-05-13
feat(translations): add translation service #575

mart-sessman:


libs/angular-components/tedi/services/translation/translation.service.ts
```diff
@@ -0,0 +1,49 @@
+import { Injectable, signal } from "@angular/core";
+import { translationsMap, TranslationMap, TediTranslationsMap } from "./translations";
+
+export type Language = "en" | "et" | "ru";
+
+@Injectable({ providedIn: "root" })
+export class TranslationService {
```
pesarv, line 7: maybe should use some kind of prefix here in naming? i would assume that a lot of applications already use ngx-translate that has same naming

libs/angular-components/tedi/services/translation/translations.ts
```diff
@@ -0,0 +1,464 @@
+import { Language } from "./translation.service";
+
+export const translationsMap = {
+  close: {
+    description: 'Used for closing',
+    components: ['CloseButton', 'Collapse', 'Notification', 'FileUpload', 'Dropdown', 'Tooltip'],
```
pesarv, line 6: whats the use for the components array? is it just for context or will it be used in future to scope translations somehow?

mart-sessman: This is just for context for developers.
I think this should be changed to optional, so when overriding or adding new translations in consumer applications, it will be not needed.

herrsiim (approved):

Issue #786, opened 2025-05-13
[useBreakpointProps]: all components that use breakpoint props result as hydration error with SSR


kasparsuvi1:
## Problem
Components that internally use `usBreakpointProps` result as hydration error with server-side rendering. Happens because itself uses `useBreakpoint` which then has default state setting with `typeof window !== 'undefined'`. And in SSR `window` object is not available.  So the `getBreakpoint` function returns null and none of the props are passed to component server-side. This results to big layout shifts as even children properties are not rendered on the server.

## Solution
* remove window usage in default state
* remove `typeof window !== 'undefined` inside useEffect as useEffect is always run on the client-side
* switch from `useEffect` to `useLayoutEffect` - which ensures that when component is rendered client-side only the first paint is done with correct props.
* Pass `defaultBreakpoint` to `useBreakpointProps` and set it default as "xs" - mobile-first approach
* Extend `BreakpointSupport` with defaultBreakpoint property and use it in every component that uses it, to allow passing different breakpoint than "xs"

Issue #778, opened 2025-05-08
what-input is automatically initalized, which will result hydration error in Next.js


kasparsuvi1:
## Problem
1. what-input package relies on client-side rendering, but currently is imported in community index file.
This makes it impossible to use Tedi components with SSR in Next.js without hydration error. Even when using `use client;` on all the component imports, because it's still rendered server-side.

2. When some application only uses Tedi components and never imports anything from Community components he doesn't get what-input features. Which will break all keyboard focus styles.

## Solution
We should manually initialize what-input as shown in the referenced link below and remove what-input import from `community/index.ts`.

Im not sure what would be the best place to initalize what-input. It should be required to use, because without it the focus styles will not pass accessibility tests. Therefore my guess would be AccessibiltyProvider, because it should  be already used in every application - from Layout component or when not using Layout AccessibiltyProvider the application should be wrapped with AccessibilyProvider.

But AccessibiltyProvider function is something else, maybe better solution is creating new component/hook and use it in the Layout and also export it separately so it can be added to applications that don't use Layout from component-library.

## Examples and information
Similar [issue ](https://github.com/ten1seven/what-input/issues/132)on what-input side. And how to initalize it manually.

PR #789, opened 2025-05-13
fix(useBreakpointProps): remove window usage in SSR, add defaultServerBreakpoint #786

kasparsuvi1:
- remove window usage in default state
- remove typeof window !== 'undefined inside useEffect as useEffect is always run on the client-side
- switch from useEffect to useLayoutEffect - which ensures that when component is rendered client-side only the first paint is done with correct props.
- Pass defaultBreakpoint to useBreakpointProps and set it default as "xs" - mobile-first approach
- Extend BreakpointSupport with defaultServerBreakpoint property and use it in every component that uses it, to allow passing different breakpoint than "xs"
- remove defaultServerBreakpoint from getCurrentBreakpointProps to avoid passing it to html elements

### DIFF
--- a/libs/react-components/src/community/helpers/hooks/use-breakpoint-props.mdx
+++ b/libs/react-components/src/community/helpers/hooks/use-breakpoint-props.mdx
@@ -29,7 +29,7 @@ interface ComponentProps extends BreakpointSupport<ComponentBreakpointProps> {
 }
 
 const Component = (props: ComponentProps) => {
-  const { getCurrentBreakpointProps } = useBreakpointProps();
+  const { getCurrentBreakpointProps } = useBreakpointProps(props.defaultServerBreakpoint);
   const { prop1, prop2, ...rest } = getCurrentBreakpointProps<ComponentProps>(props, defaultPropsObject);
 
   return <p>content</p>;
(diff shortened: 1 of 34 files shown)


kasparsuvi1:
Will add some common SSR documentation with #778

libs/react-components/src/community/helpers/hooks/use-breakpoint-props.mdx
```diff
@@ -43,3 +43,21 @@ And when using as Component:
   <p>Content</p>
 </Component>
 ```
+
+## defaultServerBreakpoint with SSR
+
+When using SSR, you can to pass the defaultServerBreakpoint prop to the component.
```
HarlesPilter, line 49: typo

libs/react-components/src/community/helpers/hooks/use-breakpoint-props.ts
```diff
@@ -19,20 +19,34 @@ export const useBreakpointProps = () => {
...
   );
 
   return { getCurrentBreakpointProps };
 };
 
+/**
+ * BreakpointSupport is a utility type that allows you to define props for different breakpoints.
+ * It extends the given type T and adds optional properties for each breakpoint except 'xs'.
+ * This is useful for creating responsive components that can have different props based on the current breakpoint.
+ * Also defaultSSRBreakpoint is added to the type, so you can set a default value for the component, when it's rendered on the server-side.
```
HarlesPilter, line 42: defaultSSRBreakpoint -> defaultServerBreakpoint

HarlesPilter (approved):

airikej (approved):

libs/react-components/src/community/helpers/hooks/use-breakpoint-props.ts
```diff
@@ -19,20 +19,34 @@ export const useBreakpointProps = () => {
     <T>(
       props: BreakpointSupport<T>,
       defaultValues: Partial<T> = {}
-    ): Omit<BreakpointSupport<T>, Exclude<Breakpoint, 'xs'>> => {
+    ): Omit<BreakpointSupport<T>, Exclude<Breakpoint, 'xs'> | 'defaultServerBreakpoint'> => {
```
mart-sessman, line 22: These sm,md,lg, xl, xxl are never used.

mart-sessman (changes requested):

libs/react-components/src/community/helpers/hooks/use-breakpoint-props.ts
mart-sessman: It causes warnings during build.

kasparsuvi1: They are needed there to not include in xs props. But added eslint-disabled comment.
(Maybe should remove all warnings from components, as it didn't appear from this PR and there is alot of more warnings.)

PR #790, opened 2025-05-13
feat(accessibiltyProvider): move initalizing what-input inside AccessibiltyProvider #778

### LINKED ISSUE #778: what-input is automatically initalized, which will result hydration error in Next.js

kasparsuvi1:
close #778

### DIFF
--- a/libs/react-components/src/tedi/helpers/hooks/use-what-input.ts
+++ b/libs/react-components/src/tedi/helpers/hooks/use-what-input.ts
@@ -1,0 +1,12 @@
+import React from 'react';
+
+/**
+ * This hook is used to load the what-input library. It is used to detect the input method used by the user (e.g. keyboard, mouse, touch).
+ * It is loaded only once when the component is mounted to avoid loading while SSR.
+ * All the applications should use this hook to load the what-input library. By default it is baked into the <Layout> component.
+ */
+export function useWhatInput() {
+  React.useEffect(() => {
+    import('what-input');
+  }, []);
+}
--- a/libs/react-components/.storybook/preview.tsx
+++ b/libs/react-components/.storybook/preview.tsx
@@ -7,7 +7,6 @@ import '../src/tedi/styles/index.scss';
 import '../src/community/styles/index.scss';
 import '../../tedi-core/src/tedi-storybook-styles.scss';
 import '../src/community/styles/storybook.scss';
-import 'what-input';
 
 export const decorators: Preview['decorators'] = [
   (Story: React.ComponentType, context: StoryContext) => {
--- a/libs/react-components/src/community/index.ts
+++ b/libs/react-components/src/community/index.ts
@@ -1,5 +1,3 @@
-import 'what-input';
-
 export * from './components/commonTypes';
 export * from './components/typography/heading/heading';
 export * from './components/typography/text/text';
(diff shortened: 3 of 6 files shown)


libs/react-components/src/tedi/helpers/hooks/use-what-input.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+/**
+ * This hook is used to load the what-input library. It is used to detect the input method used by the user (e.g. keyboard, mouse, touch).
+ * It is loaded only once when the component is mounted to avoid loading while SSR.
+ * All the applications should use this hook to load the what-input library. By default it is baked into the <Layout> component.
+ */
+export function useWhatInput() {
+  React.useEffect(() => {
+    import('what-input');
```
HarlesPilter, line 10: Wondering what this import will do when it is running twice in development when using React StrictMode

libs/react-components/src/tedi/helpers/hooks/use-what-input.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+/**
+ * This hook is used to load the what-input library. It is used to detect the input method used by the user (e.g. keyboard, mouse, touch).
+ * It is loaded only once when the component is mounted to avoid loading while SSR.
+ * All the applications should use this hook to load the what-input library. By default it is baked into the <Layout> component.
+ */
+export function useWhatInput() {
+  React.useEffect(() => {
```
HarlesPilter, line 9: Maybe using useLayoutEffect initializes it a bit sooner? Will it matter?

kasparsuvi1: I doubt, we mainly use what-input data to show focus styles with keyboard or to pass how floating should work. None of them are important at first paint.

libs/react-components/src/tedi/helpers/hooks/use-what-input.ts
kasparsuvi1: https://nodejs.org/api/modules.html#modules_caching
"Modules are cached after the first time they are loaded. This means (among other things) that every call to require('foo') will get exactly the same object returned, if it would resolve to the same file." 

ES modules should behave the same

HarlesPilter (approved):

PR #793, opened 2025-05-14
feat(breadcrumbs):angular community breadcrumb #752

herrsiim:
Breadcrumb community component.

This PR also includes 
 - Angular uptade to 19.2.11 
 - Angular router now included also in the package


libs/angular-components/tedi/services/breakpoint/breakpoint.service.ts
```diff
@@ -23,7 +25,7 @@ export type BreakpointInputsWithoutSignals<TInputs> = TInputs &
   providedIn: "root",
 })
 export class BreakpointService {
-  private currentBreakpoint = signal<Breakpoint | undefined>(undefined);
+  currentBreakpoint = signal<Breakpoint | undefined>(undefined);
```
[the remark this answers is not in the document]
herrsiim: Thank you Mr. Copilot - getter added 👍

libs/angular-components/community/components/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,102 @@
...
+   *   { label: 'Nutitelefonid', href: '/products/electronics/mobile-phones/smartphones' },
+   * ];
+   * ```
+   * @description
+   * The breadcrumbs to be displayed. Each breadcrumb should have a label and a href.
+   * The label is the text that will be displayed, and the href is the URL that the breadcrumb will link to.
+   * The breadcrumbs will be displayed in the order they are provided.
+   * The last breadcrumb will not be a link, but will be displayed as plain text.
+   */
+  crumbs = input<Breadcrumb[]>(sampleCrumbs);
```
mart-sessman, line 67: This sample data should not stay as default value. Default value should be empty array.
The description can stay for sure usage of this component.

libs/angular-components/community/components/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Used to override the breakCrumbs value to force always show singleCrumb.
+   * @default false
+   */
+  shortCrumbs = input<boolean>(false);
```
mart-sessman, line 72: What is the difference between shortCrumbs = true and singleCrumb = true?
It is confusing, can we add these 2 together into one input maybe?

mart-sessman (changes requested):

mart-sessman:
So what I am thinking is this:
We have crumbs, default value is empty list.
This component uses breakpoint service, so it's inputs can be overridden on every breakpoint.
Instead of shortCrumbs or singleCrumb inputs I would create "type" input. And type can be "single" or "multiple".

libs/angular-components/community/components/navigation/breadcrumbs/breadcrumbs.component.ts
herrsiim: Tottally forgot, removed now. 😅

libs/angular-components/community/components/navigation/breadcrumbs/breadcrumbs.component.ts
herrsiim: Good point, removed one signal.

libs/angular-components/community/components/navigation/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Breakpoint to be used for displaying single crumb.
+   * @default md
+   */
+  breakCrumbs = input<Breakpoint>("md");
```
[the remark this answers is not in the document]
herrsiim: Full breakpoint support added.

PR #821, opened 2025-05-23
feat(list): Add ability to change bullet colors #719

airikej:


libs/react-components/src/tedi/components/content/list/list.spec.tsx
```diff
@@ -52,4 +52,14 @@ describe('List Component', () => {
...
     expect(container.firstChild).toHaveClass(customClass);
   });
+
+  test('applies default bullet color class "brand" when no color is provided', () => {
+    const { container } = renderList();
+    expect(container.firstChild).toHaveClass('tedi-list--bullet-color-brand');
+  });
+
+  test('applies custom bullet color class when color prop is provided', () => {
+    const { container } = renderList({ color: 'danger' });
```
mart-sessman, line 62: Can we test all color variables here with for loop?
So every color is covered.

mart-sessman (changes requested):

mart-sessman (approved):

Issue #824, opened 2025-05-26
[Component name]:


herrsiim:
## Problem
Button component will not render text inside `<span>` element if we use ngx-translate together with http-service to display translation strings. 

During the initialization of the component, the  `{{ 'app.back' | translate }}` will render an empty string because the ngx-translate is loading the actual string information. Once the string is loaded, the ngx-translate service displays the correct string but the button's internal logic of wrapping the text string between the `<span>` elements will not run again. 
 
## Solution
Update button.component.ts lifecycle to re-run the `<span>` creation logic if the content projected into ng-content changes.

## Examples and information

![Image](https://github.com/user-attachments/assets/607e4b05-0113-442a-8606-5dddc33bab51)

PR #825, opened 2025-05-26
fix(button): Button text rendering  #824

herrsiim:


libs/angular-components/tedi/components/buttons/button/button.component.ts
```diff
@@ -56,32 +55,81 @@ export class ButtonComponent implements AfterContentInit {
...
+  }
+
+  ngOnDestroy(): void {
+    if (this.observer) {
+      this.observer.disconnect();
     }
+  }
 
-    const elements = childNodes.filter(node => node.nodeType === Node.ELEMENT_NODE);
+  private tryWrapTextNode(node: Node): void {
```
mart-sessman, line 87: Remove "try", it's misleading since there is no try-catch block inside.
"wrapTextNode" is clear enough.

mart-sessman (approved):

Issue #866, opened 2025-06-16
[Collapse]: randomUUID not found


herrsiim:
## Problem
crypto.randomUUID is not found when visiting page with http.

## Solution
Implement a fallback method for when the randomUUID() is not available

PR #867, opened 2025-06-16
fix(collapse): Collapse randomuuid not found bugfix #866

herrsiim:
Created fallback function for crypto.randomUUID if it's not available. 
 - Renamed utils folder to dev-tools. 
   - This folder should consist of tools that are related with storybook or package building.
 - Created helper folder into tedi and community folder. 
   - This folder should contain all the helper functions that tedi components use. 

This PR includes fix for Alert component imports which fixes the package building.


libs/angular-components/tedi/helpers/generateUUID.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
mart-sessman, line 1: Lets keep file naming convention like before.
generateUUID.ts --> generate-uuid.ts

Also, there is already utils folder and now helpers folder.
Maybe lets merge these two together?

libs/angular-components/tedi/index.ts
```diff
@@ -2,3 +2,4 @@ export * from "./components";
 export * from "./directives";
 export * from "./services";
 export * from "./types";
+export * from "./helpers";
```
mart-sessman, line 5: Do we need to export these or maybe let's keep it just internally?

mart-sessman (changes requested):

libs/angular-components/tedi/index.ts
herrsiim: I tried removing this, but it turns out that the ng-packager needs this, otherwise the project will not build.

mart-sessman (approved):

PR #898, opened 2025-06-26
fix(modal): remove of untagged template literals for compatibility #776

artur-langl:
remove untagged template literals from the modal for compatibility with older versions of angular

### DIFF
--- a/libs/angular-components/community/components/overlay/modal/modal.component.ts
+++ b/libs/angular-components/community/components/overlay/modal/modal.component.ts
@@ -1,5 +1,6 @@
 import {
   Component,
+  computed,
   inject,
   model,
   OnInit,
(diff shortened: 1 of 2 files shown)


libs/angular-components/community/components/overlay/modal/modal.component.ts
```diff
@@ -26,22 +29,34 @@ export type DialogData = ComponentInputs<
...
 
+  // host classes
+  readonly hostClasses = computed(() => {
+    const classes = ["tedi-modal"];
+
+    if (this.variant() === "small") {
+      classes.push("tedi-modal--small");
+    }
+
+    for (const breakpoint of Object.keys(BREAKPOINTS)) {
```
pesarv, line 52: i wouldn't take these from breakpoints since these are modal specific sized (for example there's xxl in breapoints, but no such modal max width)

pesarv (approved):

PR #905, opened 2025-07-01
feat(label-provider): export utility functions that help to create cu…

### LINKED ISSUE #904: [LabelProvider]: Types are not exposed to add own labels
## Problem
Currently the types and utility functions used to create the labelsMap are not exposed. Without them we can properly add our own new labels to the label provider. The only way is to bypass types by casting them to be `any`

## Solution
Export types and utility functions used to create the labelsMap and also find a way to allow defining new labels through module augmentation.

## Examples and information
For example when creating our own library we can't use the LabelProvider to add new labels for our components.

HarlesPilter:
closes #904


mart-sessman:
Overall I agree with adding type safety to consumer applications too, but I think current LabelProvider needs refactoring, because the setup to do this is too complicated and contains a lot of boilerplate for consumer applications.

I made my own LabelProvider, using some parts from the old one and got to this simple solution:
image

HarlesPilter:
@mart-sessman 
I think there are two main use cases when using LabelPropvider

1. Using it in an application and just overriding a few TEDI default translations
The usecase for an application was pretty simple I think, with good type safety. 
image
3. In addition to nr 1, wanting to add custom labels (Like in a wrapper library)
The example added in this MR was for this use case. To me it seems the only difference is missing `validateLabel` and generating default translations in each lang, and the LabelProvider parsing out the correct translations.

If I understand correctly then the consuming application/library has to define `description` and `components` properties as well? If we are talking just about an application that wants to add a few label overrides these should not be needed. 
What would be the type I can use to type `customLabels`.

the main point of `validateLabel` was to provider type safety for `et`, `en` and `ru` keys. So you couldn't do something like this
```
const customLabels = {
  example: {
    description: 'example',
    components: [],
    et: (condition: boolean) => condition ? 'tingimus on tõene' : 'tingimus on väär',
    en: (condition: number) => condition ? 'condition is not 0' : 'condition is 0',
    ru: () => 'random text in russian'
  }
}
```
the translation function types should always be the same.

If you found a way to keep the same level of type safety in TEDI, wrapper library and application, then I am all for a refacto. Would like to see the new code.

Issue #775, opened 2025-05-07
[Table]: create community table styles provider


pesarv:
A wrapper component to provide HTML table with TEDI styles

PR #941, opened 2025-07-16
feat(table): create community table styles provider #775

pesarv:
https://github.com/TEHIK-EE/tedi-design-system/issues/775


libs/angular-components/community/components/table/table-styles/table-styles.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="tedi-table-scroll">
```
erkidorbek, line 1: I'm not 100% sure, but if the element is scrollable, we should add cdkScrollable directive, so the events get dispatched correctly to angular.

erkidorbek:
Looks good to me despite not having cdkScrollable directive on the scrollable element.

PR #958, opened 2025-07-31
fix(form-card-error): with extra specificty, fix card error style #860

artur-langl:
image
also adds a story for a card with a error


artur-langl:
branch name doesn't fit the standard, whoops, will use the correct one next time

PR #980, opened 2025-08-19
fix(status-badge-no-status): add missing support for status badge without indicator

artur-langl:
https://artur-langl.github.io/tedi-design-system/fix/status-badge-no-status/angular/?path=/docs/community-angular-tags-statusbadge--docs


airikej:
It seems that the status bullet has some kind of white border and its still visually present with status="none"

artur-langl:
> It seems that the status bullet has some kind of white border and its still visually present with status="none"

Did not see that, will fix

artur-langl:
> It seems that the status bullet has some kind of white border and its still visually present with status="none"

Fixed it, did not properly provide null in the storybook props

airikej:
2 more things:
1. The "icon-only" badge looks like a rectangle but it should look like a square e.g in React:
image

2. There is some bug now in storybook with status="none" but with icon badge, showing this:
image

mart-sessman:
> > I think we should have one css file instead of two, even if it is copy pasted from react. Remove .react.scss file and keep the original. Otherwise seems ok to me.
> 
> done

Check your last commit, it broke story it storybook.
image

PR #1030, opened 2025-08-27
fix(floating-button): prevent text overflow with text-wrap for WCAG compliance #1001

airikej:


libs/react-components/src/tedi/components/buttons/floating-button/floating-button.module.scss
```diff
@@ -7,16 +7,23 @@ $btn-height-small: 2.5rem;
 $btn-height-large: 3rem;
 
 .tedi-floating-button {
-  box-shadow: 0 4px 10px 0 rgb(0 0 0 / 14%);
+  box-sizing: border-box;
+  text-wrap: wrap;
```
mart-sessman, line 11: Why text-wrap?

mart-sessman (approved):

